PR #4, opened 2023-03-12
Initial bevy 0.10 port

johanhelsing:
This refactors the high-level xpbd algorithm structure to take advantage of bevy_ecs schedule v3 (i.e. stageless).

The algorithm is split into two schedules, one high-level for running one high-level physics step (i.e. `DELTA_TIME`) step, and one lower-level for the inner substepping (integrate, solve etc.).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -46,7 +46,7 @@ pub struct XpbdPlugin;
 #[derive(Debug, Hash, PartialEq, Eq, Clone, ScheduleLabel)]
 pub struct XpbdSchedule;
 
-/// The substepping schedule, there number of substeps per physics step is
+/// The substepping schedule, the number of substeps per physics step is
 /// configured through the [`NumSubsteps`] resource
 #[derive(Debug, Hash, PartialEq, Eq, Clone, ScheduleLabel)]
 pub struct XpbdSubstepSchedule;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -101,7 +101,6 @@ impl Plugin for XpbdPlugin {
         );
         app.add_system(run_physics_schedule.in_set(FixedUpdateSet));
 
-        // move to plugin?
         app.add_system(
             run_substep_schedule
                 .in_set(PhysicsSet::Substeps)
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -37,7 +37,6 @@ pub type Vector = Vec3;
 
 pub const DELTA_TIME: f32 = 1.0 / 60.0;
 
-// todo: rename?
 #[derive(Debug, Hash, PartialEq, Eq, Clone, SystemSet)]
 struct FixedUpdateSet;
 
@@ -165,7 +164,6 @@ fn draw_aabbs(aabbs: Query<&ColliderAabb>, mut lines: ResMut<DebugLines>) {
 
 #[derive(Resource, Debug, Default)]
 pub struct XpbdLoop {
-    // todo: split public and private api into separate resources?
     pub(crate) accumulator: f32,
     pub(crate) queued_steps: u32,
     pub paused: bool,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -94,7 +94,7 @@ impl Plugin for XpbdPlugin {
         // move to plugin?
         app.add_system(
             run_substep_schedule
-                .in_set(PhysicsStep::Substeps)
+                .in_set(PhysicsSet::Substeps)
                 .in_schedule(XpbdSchedule),
         );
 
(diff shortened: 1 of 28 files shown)


johanhelsing:
Alright, ready for some feedback now. It's mostly about naming (see inline todos)

src/lib.rs
```diff
@@ -69,11 +79,34 @@ impl Plugin for XpbdPlugin {
...
             .register_type::<LocalCom>();
 
-        // Add stages
-        app.add_stage_before(
-            CoreStage::Update,
-            FixedUpdateStage,
-            SystemStage::parallel().with_run_criteria(run_criteria),
+        let mut xpbd_schedule = Schedule::default();
+
+        xpbd_schedule.configure_sets(
```
johanhelsing, line 83: I wonder if we should make these base sets instead of sets?

src/steps/mod.rs
```diff
@@ -8,18 +8,26 @@ pub use integrator::IntegratorPlugin;
...
     /// In the preparation step, necessary preparations and updates will be run before the rest of the physics simulation loop.
     /// For example, bevy `Transform`s are synchronized with the physics world, AABBs are updated etc.
     Prepare,
     /// During the broad phase, potential collisions will be collected into the [`BroadCollisions`] resource using simple AABB intersection checks. These will be further checked for collision in the [`PhysicsStep::NarrowPhase`].
     ///
     /// The broad phase speeds up collision detection, as the number of accurate collision checks is greatly reduced.
     BroadPhase,
+    /// Substepping is an inner loop inside a physics step, see [`PhysicsSubstep`] and [`XpbdSubstepSchedule`]
+    Substeps,
+    // todo: what about sync transforms?
```
johanhelsing: this seems to happen some way, though I don't know how... removing todo.

src/lib.rs
Jondolf: Maybe? Would it primarily make system ordering less ambiguous? My knowledge of the stageless stuff is a bit shaky currently

johanhelsing: Yeah, that's my understanding as well. Though, I think it mostly makes sense if there is a good default base set, which I don't think there is here... Just thought I'd mention it in case anyone had an opinion.

Could always revisit it later, though.

src/lib.rs
```diff
@@ -69,11 +78,34 @@ impl Plugin for XpbdPlugin {
...
+                .in_base_set(CoreSet::PreUpdate),
+        );
+        app.add_system(run_physics_schedule.in_set(FixedUpdateSet));
+
+        // move to plugin?
+        app.add_system(
+            run_substep_schedule
+                .in_set(PhysicsSet::Substeps)
+                .in_schedule(XpbdSchedule),
         );
```
Jondolf, line 109: Do you mean that we would have a plugin like `SubstepPlugin` that runs and initializes the `XpbdSubstepSchedule`? We could even move the `NumSubsteps` resource into the same file to group substepping code a bit more.

Would the substepping schedule's plugins (currently just `IntegratorPlugin` and `SolverPlugin`) be initialized in the substepping plugin as well, or should they stay in the `XpbdPlugin`?

johanhelsing: Yeah, that was the question. I started doing that, then I felt like it's kind of odd to split the overall xpbd algorithm across multiple files, and decided to keep it in `lib.rs`. I feel like when it's just adding two plugins, it's not really worth the extra indirection (adds unnecessary complexity imo).

Removing the todo now, let me know if I should do the refactor now, or we could perhaps do it in a follow-up pr.

Jondolf: Yeah I think we can keep it in `lib.rs` for now to avoid splitting the overall structure. It can always be changed later if there's a good reason.

src/lib.rs
Jondolf: We can probably leave it like this for now as there's no default base set that makes sense. If there ends up being a good reason for making them base sets, someone can open a follow-up PR.

src/steps/mod.rs
Jondolf: This is being done in the `Prepare` step, at the start of the simulation loop, as explained in the doc comment. I think originally it was being done at the end of the loop, but I think it can be done at the start as well?

johanhelsing:
XpbdSchedule:
xpbdschedule

Substep schedule:
substepschedule

Looks like there are some ambiguities, from my knowledge they were there from before, I've tried to do a pure refactor as best I could.

src/steps/mod.rs
johanhelsing: Doing it at the end would improve latency, but let's revisit it in another PR.

Jondolf:
I think we can maybe address the ambiguities in another PR, as it's not strictly related to the 0.10 migration

PR #6, opened 2023-03-12
Add basic CI setup

johanhelsing:
This is the setup I use in most of my bevy plugins.

Should make reviewing PRs a bit easier.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,73 @@
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+name: ci
+
+jobs:
+  check:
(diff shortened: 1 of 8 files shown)


johanhelsing:
See https://github.com/johanhelsing/bevy_xpbd_jondolf/pull/1 for a demo run

johanhelsing:
demo run is actually passing now.

Had to remove `bevy_audio` feature to get around alsa dependency, but removing those default feature dependencies is a good thing anyway to speed up builds and let people decide if they'd rather use bevy_kira_audio etc. for their games instead.

Jondolf:
I think this looks good, thanks!

PR #9, opened 2023-03-16
Sync physics transforms at the end of the physic schedule

johanhelsing:
This gets rid of an unnecessary 1 frame delay, where we'd always see the physics frame 1 frame behind their position in the physics world.

Also, when pausing, we'd previously never see the latest state.


Jondolf:
Thanks! I updated doc comments to reflect the changes, other than that this looks good :)

PR #11, opened 2023-03-16
Add basic integration test

johanhelsing:
Adds a very basic integration test to check that an app can run without panicking. Needs #8 and #9 to not panic (these commit are currently cherry-picked here, hence the bigger diff).

I intend to add more integration tests for determinism etc. this is just the basic structure to build on.


Jondolf:
Thanks! I agree that we should definitely add more tests at some point for things like determinism, checking if constraint values are correct etc.

PR #12, opened 2023-03-16
Fix system order ambiguities

johanhelsing:
This one depends on most of my other pending PRs. You could review them one-by-one first, the this one will make more sense :)

XpbdSchedule:
xbpd_schedule

XpbdSubstepSchedule:
substep_schedule


johanhelsing:
Rebased to clean up history (and hopefully fix the bug)

PR #14, opened 2023-03-16
More/better builder methods

johanhelsing:
This allows the builder methods to take anything that can be converted into the type of their respective fields. So if we implemented `From<nalgebra::Point2<f32>> for Pos` for instance, then we could pass points directly into `with_pos`.

Also adds missing builder methods for `lin_vel` and `ang_vel`.


Jondolf:
Looks nice!

PR #15, opened 2023-03-16
Integrator integration test

johanhelsing:
Depends on #14 

Spawn a rigid body with a velocity and check that it ends up roughly where we expect it to be.

Adds just a little bit of sanity checking.


johanhelsing:
Rebased to clean up history

PR #18, opened 2023-03-17
Add `enhanced_determinism` feature and tests

johanhelsing:
Adds [insta](https://insta.rs/) tests to ensure that we get the same output every time we run a given simulation.

Adds a new integration test that uses a copy of the cubes example, and confirms the transforms are the same each time the test is run.


johanhelsing:
Heh, I guess we just found out it's not deterministic across machines/os/something :P

johanhelsing:
Alright, determinism test is actually running *and* passing now :) Should be good to go.

PR #19, opened 2023-03-17
Use opt-level 1 in debug mode for ~100x faster examples and tests

johanhelsing:
Makes compilation slightly slower, but execution of examples tests much faster

Makes cubes_simulation_is_deterministic run in ~2 seconds instead of ~2 minutes.

Helps #18


johanhelsing:
Also, if someone checks out this repo and runs `cargo run --example cubes` it gives a slightly better impression ;)

PR #20, opened 2023-03-17
Add `f64` feature

johanhelsing:
Alternative to #7, but uses features instead of crates.


johanhelsing:
I tried having only a `f64` and not a `f32` feature, but ran into issues, there is no way to specify that you want a dependency when you don't have a feature enabled, and we need to select either `parry2d` or `parry2d-f64`.

See: https://github.com/rust-lang/cargo/issues/1839

Jondolf:
Yeah, if we opt for this feature route, I believe users would have to disable default features and manually set the features for `f64` and the desired dimension. This would perhaps make the 2D/3D crate separation kind of redundant, since you would have to specify the dimension manually anyway (for `f64`).

The original 4 crate method is a bit more complex, but it would probably be more versatile and explicit in general. I'm still not sure which one is better...

Jondolf:
Would it make sense to create a `math` module and put the math types and impls there? It would declutter `lib.rs` a bit. We could add the modules `math/single.rs` and `math/double.rs` and reimport the types in `math/mod.rs` depending on the float type.

Jondolf:
Okay, everything should be good to go now. I just fixed some dependency/feature stuff, created the `math` module and "fixed" Clippy warnings.

You can do a quick review when you have time and I'll get this merged

crates/bevy_xpbd_3d/examples/revolute_joint_3d.rs
```diff
@@ -56,10 +56,10 @@ fn setup(
 
     commands.spawn(
         RevoluteJoint::new_with_compliance(anchor, object, 0.0)
-            .with_local_anchor_1(Vec3::Y * -0.5)
-            .with_local_anchor_2(Vec3::Y * 0.5)
-            .with_aligned_axis(Vec3::Z)
-            .with_angle_limits(-0.1 * std::f32::consts::PI, 0.5 * std::f32::consts::PI),
+            .with_local_anchor_1(Vector::Y * -0.5)
```
johanhelsing, line 59: I've been thinking a bit... I'm worried that the examples become worse examples by generic like this. It's not how people would actually use the library. They'll likely just choose one precision level, `f32` or `f64` and then use those types instead of `Scalar`, `Vector` etc.. i.e. I think the library looks less ergonomic than it actually is...

Maybe it's worth duplicating at least some of the examples? They could just be feature gated.

johanhelsing:
I like how the math module turned out. Definitely a lot less clutter now 👍

crates/bevy_xpbd_3d/examples/revolute_joint_3d.rs
Jondolf: Yeah, I agree. It's better to have the examples be as simple as possible and reflect actual usage.

I'll work on this later today when I fix the errors as well (insta snapshots changed due to f64 type I think)

Jondolf: Done, I changed all the examples to use `f32` again and added it as a required feature for the examples. I didn't add any `f64` examples yet, but we can add them later if we want.

Jondolf:
Okay, _now_ everything should be working.

crates/bevy_xpbd_3d/examples/revolute_joint_3d.rs
Jondolf: Okay I added `cubes_f64` and it seems to work as expected.

PR #21, opened 2023-03-17
`f32` and `f64` additive features

johanhelsing:
Another option. I wanted to see what it would take to have both `f32` and `f64` function at the same time.

This makes each implementation available under `bevy_xpbd_3d::single::XpbdPlugin` and `bevy_xpbd_3d::double::XpbdPlugin`. If only one of them is defined, adds re-exports at the crate level, i.e. `bevy_xpbd_3d::XpbdPlugin`.

Not sure what I think of it though, but just thought I'd share.


Jondolf:
Interesting idea, haven't seen it before. This does maybe feel a bit hacky though, and seems to basically accomplish the same thing as making separate crates for `f32` and `f64`, except with features and modules

johanhelsing:
Yeah, definitely feels hacky.

Another option would be to use generics, so `XpbdPlugin<f32>` for instance, and then use associated types for everything (`XpbdPlugin<f32>::Vector` would be `Vec3`)... if it's even doable it would probably be a lot less readable, and everything would probably have to go into impl<T> blocks.

PR #23, opened 2023-03-23
Fix default 3D inertia

Jondolf:
This changes the default 3D inertia tensor value to be zero instead of an identity matrix. A body should have no mass or inertia by default.

Previously, when adding a collider with its own mass properties, the `Inertia` of the body was slightly too high due to the body having some initial inertia given by the identity matrix. This could be noticed in the examples by removing the collider mass properties and setting the body's own mass properties instead; the inertia was different when comparing these two methods of adding mass. This should now be fixed.


johanhelsing:
Nice catch

PR #24, opened 2023-03-25
Don't correct for restitution on normal_vel < 0

johanhelsing:
ref: https://discord.com/channels/691052431525675048/1084206740570124372/1089161165088772146

This check doesn't appear in the paper, but without it, I'm getting issues on object with sharp corners (in my own version of bevy_xpbd).

https://user-images.githubusercontent.com/2620557/227721679-12c3d16f-08ed-4ace-a88a-4c37f3cc09dc.mp4

I think the alternative would be to say that perfectly sharp corners are just not allowed.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -53,6 +53,12 @@ pub(crate) fn get_restitution(
     gravity: Vector,
     sub_dt: Scalar,
 ) -> Vector {
+    if normal_vel < 0.0 {
+        // if bodies are moving away from each other, they should be able to do
+        // so unconstrained by restitution corrections.
+        return Vector::ZERO;
+    }
+
     let mut restitution_coefficient = (restitution1.0 + restitution2.0) * 0.5;
 
     // If normal velocity is small enough, use restitution of 0 to avoid jittering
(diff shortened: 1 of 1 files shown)


Jondolf:
This could fix that particular problem, but it seems like it causes a ton of explosions and jitteriness (e.g. in the "cubes" example)

johanhelsing:
Ouch, ok... wonder why that is?

Jondolf:
It could be a 3D problem though, as trying a similar example in 2D with squares seems to work better. Though I didn't notice this actually fixing any collision issues in 2D either...

johanhelsing:
Alright, back to the drawing board, i suppose...

Aceeri:
@johanhelsing Are you sure this is a bug? I'd assume rubbing 2 objects would pull the object with it based on friction. Or are we talking about the initial slide not doing anything?

johanhelsing:
Yeah, friction is set to zero here. Direction change comes from restitution.

Jondolf:
I'll close this for being stale and probably outdated.

PR #25, opened 2023-04-01
Add `CoefficientCombine` rule for `Restitution` and `Friction`

Jondolf:
This adds a `CoefficientCombine` enum with the `Average`, `Min`, `Multiply` and `Max` variants, where `Average` has the lowest priority, and `Max` has the highest priority (in case of conflicting combine rules). The default is still `Average` like before.

I would have also added something like `Custom(fn(a: Scalar, b: Scalar) -> Scalar)`, but it would cause issues with `Copy` and `Reflect`, so I left it out.

I also added a couple builder methods and a test for the combine rules.

Combining two `Restitution` components:

```rs
let restitution1 = Restitution::new(0.3).with_combine_rule(CoefficientCombine::Average);
let restitution2 = Restitution::new(0.7).with_combine_rule(CoefficientCombine::Multiply);
let result = Restitution::new(0.21).with_combine_rule(CoefficientCombine::Multiply);

assert_eq!(restitution1.combine(restitution2), result);
```

Friction works similarly, but the same combine rule is currently used for both static and dynamic friction.

### DIFF
--- a/crates/bevy_xpbd_2d/examples/move_marbles.rs
+++ b/crates/bevy_xpbd_2d/examples/move_marbles.rs
@@ -106,10 +106,7 @@ fn setup(
                     },
                     ..default()
                 },
-                RigidBodyBundle {
-                    restitution: Restitution(0.3),
-                    ..RigidBodyBundle::new_dynamic().with_pos(pos)
-                },
+                RigidBodyBundle::new_dynamic().with_pos(pos),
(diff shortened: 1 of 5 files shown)


Jondolf:
Huh, weird that CI is failing on `cubes_f64`. It works fine for me, and I didn't change anything about the CI or the example...

PR #26, opened 2023-04-02
Update GitHub Actions CI

Jondolf:
```
Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2, actions-rs/toolchain@v1, actions-rs/cargo@v1.
```

I updated to `actions/checkout@v3` and replaced the actions-rs commands with manual commands because actions-rs is [no longer maintained](https://github.com/actions-rs/toolchain/issues/216).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,63 +12,44 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout sources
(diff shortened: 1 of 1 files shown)


Jondolf:
Also, the CI test suite is currently failing on `cubes_f64` for some reason, but it's not caused by this update.

PR #30, opened 2023-06-01
Allow all examples to run in f32/f64 depending on feature flags

Aceeri:
Removes the separation of cubes_f64 and the rest of the examples where it should just work to compile with the corresponding feature flag.


Jondolf:
Thanks! I think the actual code looks good. Adding bevy_editor_pls here seems a bit random though, so I'm wondering if it's unintenional or for debug purposes? If we want to add it to some examples, another PR might be better.

Also, I'm just wondering if we actually want to keep `f32` and `f64` separate, since it was originally kind of a deliberate choice. In most applications, people would just use either `f32` or `f64`, so supporting both in the examples doesn't reflect common usage quite as well. I'm worried people will look at the examples and see custom stuff like `Scalar`, `Quaternion` and `.as_f32()`, and assume that they have to be used instead of just `f32` and `Quat`. For new people it might make the API seem more cumbersome than it is.

I'm not sure if this is an actual problem though, and I guess I could just add some kind of "Running the examples" section in the README.md and explain that the examples support both precisions, which is why special types are used.

Let me know if you have any thoughts on this, I haven't decided anything yet

Aceeri:
> Thanks! I think the actual code looks good. Adding bevy_editor_pls here seems a bit random though, so I'm wondering if it's unintenional or for debug purposes? If we want to add it to some examples, another PR might be better.
> 
> Also, I'm just wondering if we actually want to keep `f32` and `f64` separate, since it was originally kind of a deliberate choice. In most applications, people would just use either `f32` or `f64`, so supporting both in the examples doesn't reflect common usage quite as well. I'm worried people will look at the examples and see custom stuff like `Scalar`, `Quaternion` and `.as_f32()`, and assume that they have to be used instead of just `f32` and `Quat`. For new people it might make the API seem more cumbersome than it is.
> 
> I'm not sure if this is an actual problem though, and I guess I could just add some kind of "Running the examples" section in the README.md and explain that the examples support both precisions, which is why special types are used.
> 
> Let me know if you have any thoughts on this, I haven't decided anything yet

The bevy_editor_pls stuff was just for debugging the changes a bit better and trying to figure out the instability issues, will remove it.

I'd say just explaining that examples support both precisions would be useful for less maintenance. I was mostly thinking it'd be nice to be able to switch between f32/f64 with just the feature flag, but I realize now that wouldn't really be the case...

Jondolf:
I think we can support both precisions in the examples and just mention it in the README.md. It's probably the easiest and most sensible way for now. If people find it confusing, we can always just revert to using one precision.

The constructor generics method won't work anymore since I removed the bundles in favor of a simpler and more ergonomic API (see #37), but I guess we could implement `From<Vec3>` and `From<DVec3>` etc. for the components so you could do e.g. `Pos::from(Vec3::X)` in both precisions. This might be more confusing though so I think it's fine to use the generic types like `Vector` in the examples as long as it's explained.

Would it be fine if I made another PR for this in a couple of days? I'm probably going to update some of the examples to be simper and cleaner before release, so it's probably not worth fixing all of the merge conflicts that are on this branch. I like the `AdjustPrecision` and `AsF32` traits though so I'll add you as a co-author for that

Aceeri:
Ya thats fine! :)

Jondolf:
Added `AdjustPrecision` and `AsF32` in fbcea35, closing this PR in favor of #48.

PR #32, opened 2023-06-08
Implement sleeping

### LINKED ISSUE #2: Unstable collisions
You mentioned in #1 that you were struggling with "unstable collisions", and this was something you'd like to fix before releasing. Just though it might be useful to have an issue for it.

Maybe explaining it and having some extra eyes on it would help :)

(fwiw I was also having trouble with unstable things as I added friction in the tutorial series), and I never quite figured it out, perhaps this is the same thing?

Jondolf:
Depends on #31.

## Description

Implements sleeping to improve performance and reduce small jitter.

Bodies with the `Sleeping` component are not simulated by the physics engine. Bodies are marked with the component when their linear and angular velocities are below the values of the `SleepingThreshold` resource for a duration indicated by the `DeactivationTime` resource.

The `TimeSleeping` component is used to track the time that the velocity has been below the threshold, and it is reset to 0 whenever the threshold is exceeded.

Bodies are woken up when they interact with active dynamic or kinematic bodies or joints, or when gravity changes, or when the body's position, rotation, velocity, or external forces are changed.

Sleeping can be disabled for all bodies with a negative sleeping threshold, and for specific entities with the `SleepingDisabled` component.

## Examples

Below is the current behaviour in a couple examples. The red color indicates that a body is sleeping and isn't being simulated.

First, the `cubes` example. At the end I change the velocity of the boxes. Sleeping doesn't help keep the cube stack stable unless a really high sleeping threshold is specified, but it does help keep them still on the ground.

https://github.com/Jondolf/bevy_xpbd/assets/57632562/63d8948c-6ff0-4665-9b3f-9066c087e665

`move_marbles` example:

https://github.com/Jondolf/bevy_xpbd/assets/57632562/bf19fc03-1538-4fc4-acb7-611cf831891e

## Considerations and future improvements

- As can be seen in the `cubes` example, sleeping doesn't fix jittery collisions (#2). However, it does help keep slowly drifting bodies still, so in many cases it might be enough until a proper fix for unstable collisions is found.
- The default `SleepingThreshold` should perhaps be scaled depending on the world scale, e.g. in 2D, objects might be hundreds of pixels wide, so a threshold of 0.1 would be tiny. Maybe we should introduce a `PhysicsScale` resource, similar to rapier?
- Sleeping can cause unrealistic behaviour in some cases, e.g. removing the floor under a sleeping body will leave the body floating. But AFAIK this is a thing in other engines as well, and it should probably just be mentioned in the docs somewhere.
- Currently, sleeping is done per body. Many physics engines seem to use an *island architecture* where entities are grouped into islands, and sleeping controls the entire islands. This might have some benefits that are worth considering later.

## Todo

- [x] Wake up bodies when their position or rotation is changed (constraints waking up bodies was bugged, now things work)
- [x] Figure out better place for sleeping systems than the sync step
  - I added a separate step and plugin for sleeping since I couldn't come up with another logical place.
- [x] Allow disabling sleeping for specific entities using a component
- [x] Allow separate sleeping threshold for linear and angular velocity
- [x] Add some more documentation


Jondolf:
Everything should be working and documented now. I'll merge this, feel free to open issues if you encounter any problems.

PR #77, opened 2023-07-16
Move `SpatialQuery` methods to `SpatialQueryPipeline`

LeshaInc:
This allows using spatial queries outside systems, for example in `AsyncComputePool`.

`SpatialQuery` methods remain, but now they delegate to `SpatialQueryPipeline`.

## Future work

After this change I'm wondering if `SpatialQuery` is even necessary. It uses `ResMut` internally, which means systems performing queries can't run in parallel. The only method taking `&mut self` is `update_pipeline`. Thus, I propose removing `SpatialQuery` and adding `update_spatial_query_pipeline` system, which would be added to `PostUpdate` by default (and additional copies could be added by users). All user code would have to change `SpatialQuery` to `Res<SpatialQueryPipeline>`.

### DIFF
--- a/src/plugins/spatial_query/pipeline.rs
+++ b/src/plugins/spatial_query/pipeline.rs
@@ -4,7 +4,16 @@ use crate::prelude::*;
 use bevy::{prelude::*, utils::HashMap};
 use parry::{
     partitioning::{Qbvh, QbvhUpdateWorkspace},
-    query::{DefaultQueryDispatcher, QueryDispatcher},
+    query::{
+        details::{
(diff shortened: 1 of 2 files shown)


Jondolf:
Thanks for implementing this, it's definitely overall a better and more flexible approach that should be very beneficial for more advanced use cases.

I also agree that it would probably be better to remove the `SpatialQuery` system parameter like you described. I'm not sure what the name of the resource should be though, since to me "pipeline" seems more like an internal thing that isn't for common usage; maybe just `Res<SpatialQueries>`? That can be done in a follow-up PR though.

LeshaInc:
`SpatialQueries` sounds good.

PR #78, opened 2023-07-16
Add `enabled` flag to `PhysicsDebugConfig` to globally disable debug rendering at runtime

LeshaInc:
This allows easily turning debug renderer on and off by changing `enabled` field of `PhysicsDebugConfig`

Also it matches similar approach by `GizmoConfig` from bevy and `DebugRenderContext` from rapier.

### DIFF
--- a/src/plugins/debug/configuration.rs
+++ b/src/plugins/debug/configuration.rs
@@ -62,10 +62,12 @@ impl PhysicsDebugConfig {
         }
     }
 
-    /// Disables all debug rendering.
+    /// Creates a [`PhysicsDebugConfig`] configuration with debug rendering enabled but all options turned off.
+    ///
+    /// Note: this doesn't affect entities with [`DebugRender`] component; their debug gizmos will be visible.
     pub fn none() -> Self {
         Self {
-            enabled: false,
+            enabled: true,
             axis_lengths: None,
             aabb_color: None,
             collider_color: None,
--- a/src/plugins/debug/configuration.rs
+++ b/src/plugins/debug/configuration.rs
@@ -7,6 +7,8 @@ use bevy::prelude::*;
 #[derive(Reflect, Resource)]
 #[reflect(Resource)]
 pub struct PhysicsDebugConfig {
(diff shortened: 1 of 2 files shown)


src/plugins/debug/configuration.rs
```diff
@@ -58,9 +62,10 @@ impl PhysicsDebugConfig {
         }
     }
 
-    /// Disables all debug rendering for this entity.
+    /// Disables all debug rendering.
     pub fn none() -> Self {
         Self {
+            enabled: false,
```
Jondolf, line 68: This feels like a potential footgun, since users might expect e.g. `PhysicsDebugConfig::none().with_collider_color(Color::WHITE)` to work.

I'm not sure what the best approach would be, but maybe we could just use `enabled: true` here, but clarify the doc comment to be:

```rust
/// Creates a [`PhysicsDebugConfig`] configuration with debug rendering enabled but all options turned off.
```

or something like that.

Jondolf:
I wonder if this option should also disable debug rendering for entities with the `DebugRender` component. It might be useful to be able to disable *all* debug rendering without having to remove the components.

LeshaInc:
> I wonder if this option should also disable debug rendering for entities with the `DebugRender` component. It might be useful to be able to disable _all_ debug rendering without having to remove the components.

Doesn't it already disable all systems inside the plugin?

Jondolf:
> > I wonder if this option should also disable debug rendering for entities with the `DebugRender` component. It might be useful to be able to disable _all_ debug rendering without having to remove the components.
> 
> Doesn't it already disable all systems inside the plugin?

Oh my bad, it uses a run condition so it doesn't run the systems at all. For some reason I thought it still ran it for entities with `DebugRender` and skipped the rest, not sure why. But you're right👍

src/plugins/debug/configuration.rs
Jondolf: Looks good now, thanks

Jondolf:
Looks good, I can merge this now

PR #84, opened 2023-07-17
 Improve simulation stability when object are far from world origin 

LeshaInc:
During integration and constraint solving, position gets incremented by very small amounts, especially at high substep count.  When the objects are far from origin, small increments to position are rounded to zero. To mitigate that, I've added `AccumulatedTranslation` component, which gets incremented instead of `Position`. Then, at the end of each substep, a system updates the actual position.

This change seems to be the magic bullet for "far from origin" kinds of problems. But it's not perfect yet.

Here's the cubes example with origin set to (10000, 10000). When using `f32`, the next float after `10000` is `10000.0009766`, which gives precision of roughly `1e-3` . 

Here's the before:

https://github.com/Jondolf/bevy_xpbd/assets/42153076/3e96809b-8cfa-4fc2-b288-05c8a2c4f37e

Here's the after:

https://github.com/Jondolf/bevy_xpbd/assets/42153076/b6a717d5-a981-428a-9e44-f1976f89ef50

As you can see, it starts very promising, but then explodes. I don't know what causes that, but If I change substeps from 12 to 6 (not part of PR), it doesn't explode:

https://github.com/Jondolf/bevy_xpbd/assets/42153076/6b7ecf5a-7824-4c91-b933-4bca0778304f

Also, I've changed contacts to be queried in local space, since that's what parry does internally, only then to transform it back to world space (and back to local in `PenetrationConstraint`). This improves precision.

Additionally, I've found a few places where values of different orders of magnitude were subtracted (for example when computing relative motion of contact points): I've reordered operands to try and keep relative magnitudes close.

# Future work

In the future, perhaps it would be beneficial to update actual position not on every substep, but on every _step_ instead. This way, even on very large substep counts (where dt is very tiny) everything should work fine.

### DIFF
--- a/crates/bevy_xpbd_3d/examples/cubes.rs
+++ b/crates/bevy_xpbd_3d/examples/cubes.rs
@@ -22,6 +22,9 @@ fn setup(
     mut materials: ResMut<Assets<StandardMaterial>>,
     mut meshes: ResMut<Assets<Mesh>>,
 ) {
+    let origin = Vector::new(0.0, 0.0, 0.0);
+    let density = 1.0;
+
     let cube_mesh = meshes.add(Mesh::from(shape::Cube { size: 1.0 }));
 
     // Ground
@@ -33,7 +36,7 @@ fn setup(
             ..default()
         },
         RigidBody::Static,
-        Position(Vector::NEG_Y * 2.0),
+        Position(origin + Vector::NEG_Y * 2.0),
         Collider::cuboid(100.0, 1.0, 100.0),
     ));
 
@@ -48,6 +51,9 @@ fn setup(
                     y as Scalar * (cube_size + 0.05),
                     z as Scalar * (cube_size + 0.05),
                 );
+
+                let collider = Collider::cuboid(cube_size, cube_size, cube_size);
+
                 commands.spawn((
                     PbrBundle {
                         mesh: cube_mesh.clone(),
@@ -56,8 +62,9 @@ fn setup(
                         ..default()
                     },
                     RigidBody::Dynamic,
-                    Position(pos + Vector::Y * 5.0),
-                    Collider::cuboid(cube_size, cube_size, cube_size),
+                    ColliderMassProperties::new_computed(&collider, density),
+                    collider,
+                    Position(origin + pos + Vector::Y * 5.0),
                     Cube,
                 ));
             }
@@ -77,8 +84,8 @@ fn setup(
 
     // Camera
     commands.spawn(Camera3dBundle {
-        transform: Transform::from_translation(Vec3::new(0.0, 12.0, 40.0))
-            .looking_at(Vec3::Y * 5.0, Vec3::Y),
+        transform: Transform::from_translation(origin + Vec3::new(0.0, 12.0, 40.0))
+            .looking_at(origin + Vec3::Y * 5.0, Vec3::Y),
         ..default()
     });
 }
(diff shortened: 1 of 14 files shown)


LeshaInc:
After merging #83 cubes example with origin set to (10000, 10000) still explodes, but after 10 seconds instead of 4.

https://github.com/Jondolf/bevy_xpbd/assets/42153076/492693d7-7f7e-49d6-9430-dca1ce83527a

I've tried increasing AABBS like so:
```
mins -= half_extents * 0.1;
maxs += half_extents * 0.1;
```
I've waited for a few minutes and so no explosion, and so I think the problem is caused by missed collision pair after some amount of substeps, since the broad phase is run only once.

Do you have any clues regarding this explosive behavior?

Jondolf:
I'm still getting explosions with AABBs expanded further like that, although I'm not getting the issue when the cubes are initially stacked (we might have different OSs though, I'm on Windows).

In your case I'm guessing all the cubes were marked as sleeping, so they weren't simulated and couldn't explode even if the issue is still there. If I keep moving the cubes frequently, I still get explosions:

https://github.com/Jondolf/bevy_xpbd/assets/57632562/e2d7bfbb-4702-4c19-912f-4a3e4de7d70d

I don't know what's causing the explosive behaviour, but I think a couple things are implemented slightly incorrectly.

1. I'm pretty sure constraints and `solve_vel` should use `local_r1`/`local_r2` rotated to match the current pose instead of `world_r1`/`world_r2`. Using the original contact points is wrong, since the body is often rotated during the constraint solver. When I previously tried this, I was getting mixed results, so I'm not entirely sure.
2. Dynamic friction is wrong, since it causes heavier bodies to come to a stop earlier. See #52.
3. Static friction also feels slightly wrong, but I'm not entirely sure without more testing. I have found several slightly different implementations and I'm not sure which one is correct.
4. Probably unrelated, but currently the narrow phase is inside the constraint solver instead of a separate step. Once a collision is detected, the constraint is solved immediately, instead of computing all collisions beforehand. The current implementation seems to be more stable, but it would be nice to have a separate narrow phase for parallelism and modularity.

These are probably not directly related, since disabling friction didn't seem to help when I tried a while ago, and the other changes cause mixed results. Using f64 seems to help a lot, so I suppose it could be another precision issue somewhere.

Jondolf:
As for this PR, it's definitely a massive improvement when using high masses, small dt or a world origin that is far away.

Overall, I l like the `AccumulatedTranslation` approach, although it would be nice to have more docs regarding what it's used for and why it exists, at least for other contributors.

It might also make sense to change it to `pub(crate)` or to remove it from the prelude, since I consider it to be an implementation detail of the solver, and users will almost never need to use it directly.

LeshaInc:
> It might also make sense to change it to `pub(crate)` or to remove it from the prelude, since I consider it to be an implementation detail of the solver, and users will almost never need to use it directly.

It should be public for people who want to implement their own constraints.

Jondolf:
Yeah fair, it should be public as long as it's clear what it's used for.

LeshaInc:
I tried removing `AccumulatedTranslation` from prelude, but it turned out to be very cumbersome due to glob imports everywhere.

Jondolf:
It would need to be outside of `components`, maybe in `solver` or something. But it's probably fine in the prelude for now. The docs are good so I think I can merge this.

I would maybe change the cubes example back to what it was, since the origin is unnecessary for the actual example. Could eventually add a separate example with a very far away origin and/or high mass ratios though. Also FYI, you can convert a `Vector` to f32 using `.as_f32()`.

Issue #85, opened 2023-07-18
Cubes start moving after a complete stop


LeshaInc:
Cubes start moving after a complete stop, as if they are possessed:

https://github.com/Jondolf/bevy_xpbd/assets/42153076/a26206d8-553e-421f-8d9e-cb8c89c3858d

The issue is mitigated by enabling `f64` precision.

Note that it happens even when objects are close to (0, 0) and is unrelated to #84: I've tested both with and without the PR, same result.

Jondolf:
Related to #2 (the end), but this one is more up to date and focused on the drifting

Jondolf:
This is fixed by #90 in terms of collisions against convex meshes like the ground in that video, but I had some problems with non-convex collisions, so there's a separate issue for that: #93

PR #90, opened 2023-07-19
Use contact manifolds instead of single contacts for collisions

Jondolf:
Currently, each collision only uses one contact. When a box is on the ground, the contact point oscillates between the corners, which can often cause significant instability and positional drift.

This PR changes this behaviour by computing multiple contact points using contact manifolds. This allows all four corners of a cube to be in contact at the same time, which essentially eliminates drifting, closing #85.

## Comparison

Before: 

https://github.com/Jondolf/bevy_xpbd/assets/57632562/5aeaa265-ce15-465b-9eab-42a732a1f657

After:

https://github.com/Jondolf/bevy_xpbd/assets/57632562/a4136550-0872-442c-ad36-cbf4d105e1d0

As you can see, the cubes now land in a much more stable way, drifting is almost nonexistent, and explosions or small jumps are significantly reduced.

### Small jumps

Edit: Fixed this for convex colliders, but non-convex colliders are still buggy. Read my comment below this post.

Despite drifting being essentially removed and explosions being heavily reduced, there are still occasional jumps, which might be because the constraint only uses the initial penetration depths (from Parry) instead of computing them based on the current state, which would cause the applied correction to be too large.

This seems to be mitigated by manually computing the penetration depth based on contact points computed at the current state, just like the XPBD paper recommends:

```rust
let p1 = body1.position.0
    + body1.accumulated_translation.0
    + body1.rotation.rotate(self.local_r1);
let p2 = body2.position.0
    + body2.accumulated_translation.0
    + body2.rotation.rotate(self.local_r2);
let penetration = (p1 - p2).dot(normal);
```

However, this causes problems with trimesh colliders, so I didn't add it yet. This will have to be investigated later.

### Non-convex colliders

These changes only fix convex-convex collisions, and non-convex colliders (like trimeshes) still have the old behaviour. Read my comment below this post.

### Performance

As you might expect, using multiple contact points instead of one can be more expensive, but the difference seems to actually be quite small.

Parry supports reusing the previous contact manifold to speed up the computation, so it could even be faster than the old method if implemented correctly. I didn't implement this yet due to facing more problems, but it will have to be tried again, as it should give a significant performance boost.

### API change

Collisions are no longer just singular contacts, so I added a `Contacts` struct for a collision between two entities. It contains the entities and a `Vec<ContactManifold>`, where each contact manifold contains the entities, a normal, and the contacts, which are represented as `Vec<Contact>`.

This is quite a few added layers, so it would be nice to add simple methods to e.g. get all contacts as a flat structure or to compute the deepest contact.

### DIFF
--- a/crates/bevy_xpbd_3d/examples/basic_kinematic_character.rs
+++ b/crates/bevy_xpbd_3d/examples/basic_kinematic_character.rs
@@ -134,14 +134,18 @@ fn kinematic_collision(
     mut bodies: Query<(&RigidBody, &mut Position)>,
 ) {
     // Iterate through collisions and move the kinematic body to resolve penetration
-    for Collision(contact) in collision_event_reader.iter() {
+    for Collision(contacts) in collision_event_reader.iter() {
         if let Ok([(rb1, mut position1), (rb2, mut position2)]) =
-            bodies.get_many_mut([contact.entity1, contact.entity2])
+            bodies.get_many_mut([contacts.entity1, contacts.entity2])
         {
-            if rb1.is_kinematic() && !rb2.is_kinematic() {
-                position1.0 -= contact.normal * contact.penetration;
-            } else if rb2.is_kinematic() && !rb1.is_kinematic() {
-                position2.0 += contact.normal * contact.penetration;
+            for manifold in contacts.manifolds.iter() {
+                for contact in manifold.contacts.iter() {
+                    if rb1.is_kinematic() && !rb2.is_kinematic() {
+                        position1.0 -= contact.normal * contact.penetration;
+                    } else if rb2.is_kinematic() && !rb1.is_kinematic() {
+                        position2.0 += contact.normal * contact.penetration;
+                    }
+                }
             }
         }
     }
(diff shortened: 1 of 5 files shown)


Jondolf:
I made a few more changes. I noticed that this causes convex meshes (like the cubes) to sink into non-convex meshes (like trimeshes), so I had to revert it so that collisions involving non-convex meshes keep the old single contact point behaviour. This means that collisions against trimeshes will still be a bit buggy for now, but collisions between convex shapes should be much better. I will open an issue for this later.

Additionally, I changed the penetration constraint to compute the penetration depth for convex colliders based on the current state like I described in my previous message. This seems to eliminate pretty much *all* of the remaining explosiveness, as I'm no longer seeing any jumping cubes. However, for non-convex colliders the problem still persists as I said.

Finally, I also changed the computation of r1 and r2 to take into account the current pose of the bodies instead of using the original world_r1 and world_r2. I'm not 100% sure if this is how it should be implemented, but other people and the original paper seem to suggest so. I'm not seeing a noticeable difference, but I believe it's now correct.

I'll probably merge these changes this evening and open an issue for the non-convex collider case.

PR #92, opened 2023-07-20
Setup `PhysicsSchedule` and `SubstepSchedule` to use single-threaded executor

LeshaInc:
By default schedules use multi-threaded executor, but since all systems are chained together and run one after another, using the multi-threaded executor is pointless and adds overhead.

This improves performance in benchmarks by between 12% and 50%:

3x3 cubes, 30 steps:
image

5x5 cubes, 30 steps:
image

10x10 cubes, 30 steps:
image

The relative speedup is smaller with more cubes, since task overhead is static.

In the future, systems should be internally parallelized (probably using rayon), so using the single threaded executor will free other threads for performing parallel work.

### DIFF
--- a/src/plugins/setup.rs
+++ b/src/plugins/setup.rs
@@ -2,7 +2,10 @@
 //!
 //! See [`PhysicsSetupPlugin`].
 
-use bevy::transform::TransformSystem;
+use bevy::{
+    ecs::schedule::{ExecutorKind, ScheduleBuildSettings},
+    transform::TransformSystem,
+};
 
 use crate::prelude::*;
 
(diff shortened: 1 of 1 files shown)


Jondolf:
Thanks, this definitely seems like a pretty substantial decrease in overhead. I appreciate the benchmarks!

To be clear, not *all* systems are chained together and some of the chaining could maybe be removed, but these are mostly very cheap systems, so having them run in parallel is probably not worth the overhead.

Collision detection and the solver should definitely be parallelized in the future. For the solver, we would probably need to form some kind of constraint graph and use e.g. red-black graph coloring to find partitions that can be run in parallel. That's a problem for the future though, and it will probably require quite a bit of work and research

Issue #16, opened 2023-03-16
Remove `Pos` and `Rot` in favor of bevy's `Transform`


Jondolf:
Currently, `bevy_xpbd` uses the components `Pos` and `Rot` for describing the position and orientation of bodies. At the end of each iteration of the main physics loop, these are synchronized with bevy's `Transform`s.

I'd like to discuss if we should remove the separate components and move to just using bevy's own `Transform`s.

## Advantages

- More familiar and convenient for bevy users
- More complete APIs
- Reduces code and maintenance overhead (for example, no custom `Rot` implementations for 2D and 3D, all rotations use `Quat`s)
- Less components and nicer queries (`Transform` and `PrevTransform` instead of `Pos`, `PrevPos`, `Rot`, and `PrevRot`)
- No need for synchronization step
- If/when bevy gets an editor eventually, it would make more sense to manipulate `Transform`s directly
- It's what most physics engines seem to do (like `bevy_rapier`)

## Disadvantages

- `Transform` hierarchies, scaling and shearing can cause serious problems (see comments below)
- Parallellism might be harder, since two systems can't access the same `Transform` at once. Separate systems with `Pos` and `Rot` can run simultaneously.
- No `f64` version of `Transform` unless we make a custom one (adding `f64` as a crate feature: #7)
- In 2D, `Pos` uses `Vec2` and `Rot` uses a scalar value. `Transform`s just have a `Vec3` and a `Quat`, which can be considered less ergonomic for 2D. However, bevy users are used to working with them, and they might provide more freedom, so this may not be an issue.
- Migration pain (not a problem for us now, but could be in the future if this crate becomes more popular)
- Something else?

## Conclusion

Personally, at least from an API standpoint, I think it makes sense to move to using just `Transform`s, as it would generally be much more convenient and familiar for bevy users, and it would also probably reduce the engine's code and complexity. Using `Transform`s for physics can cause many problems as well though, which is why I'd like to hear what others think.

Aceeri:
https://github.com/dimforge/bevy_rapier/pull/219#issuecomment-1194383319
Some discussion on this from @HackerFoo

HackerFoo:
Physics uses isometries (no scale or shear), and these must be global (from the physics engine's point of view), so it's not a good match for `Transform`. Implementing `From` / `Into` makes it easy to convert between transform types.

HackerFoo:
[...]

PR #96, opened 2023-07-25
Support `Transform` for moving and positioning bodies

Jondolf:
Currently, the only way to move and position bodies after initialization is to use the `Position` and `Rotation` components. Using these separate components has many benefits over using `Transform` for physics, as discussed in #16.

However, if we want Bevy XPBD to feel as integrated into Bevy as possible, it would make sense to allow using `Transform` directly for moving bodies, as it is what Bevy and the rest of its ecosystem typically uses. This would also allow third party crates to move bodies in a more backend-agnostic way without having to have explicit support for Bevy XPBD.

This PR adds a new synchronization system for `SyncPlugin` that updates `Position` and `Rotation` based on changes in `GlobalTransform`. This does *not* replace the separate components, but rather adds another way to position bodies in a more familiar and native way.

The new behaviour is entirely optional, and there is a new `SyncConfig` resource for disabling syncing features.

### ~~`Transform` changes are only handled if they happen before physics~~ (fixed)

**Update**: I fixed this by changing the `GlobalTransform` change detection a bit and running transform propagation before *and* after each physics frame. `Transform`s can now be modified at any time, and everything should work normally. The problem described below no longer applies.

To only detect user changes in `GlobalTransform`, I use a `PreviousGlobalTransform` component that stores the global transform from the very beginning of each frame. If the transform has changed before physics is run, `transform_to_position` is run.

This avoids unnecessary runs of the system and circumvents some change detection problems, but it also means that `Transform` updates are only handled if they happen before physics is run.

### Additional transform propagation

To handle hierarchies and to have `GlobalTransform` change detection working properly, I need to run a total of three copies of Bevy's transform propagation: one before physics, one right after physics, and a final one after `position_to_transform` syncing.

I don't think it's particularly expensive, but it would be nice to find a way to avoid running it so many times. However, I'm not sure if it's avoidable if we want to handle the `Transform` syncing in a robust manner.

### Best practices for positioning bodies

There are now two ways to position bodies, either using `Transform` or using `Position` and `Rotation`. The best practises are unclear here; should people use transforms, which is considered to be a more native way, or should they continue to use `Position` and `Rotation`, which is what the engine itself uses?

For now, I think it's enough to just mention both ways in the docs and let people make their own conclusions. Later on, we can maybe make clearer recommendations, and maybe even revisit #16 in some capacity.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -468,8 +468,8 @@ pub struct SubstepSchedule;
 /// 1. `Prepare`: Responsible for initializing [rigid bodies](RigidBody) and [colliders](Collider) and
 /// updating several components.
 /// 2. `StepSimulation`: Responsible for advancing the simulation by running the steps in [`PhysicsStepSet`].
-/// 3. `Sync`: Responsible for synchronizing physics components with other data, like writing [`Position`]
-/// and [`Rotation`] components to `Transform`s.
+/// 3. `Sync`: Responsible for synchronizing physics components with other data, like keeping [`Position`]
+/// and [`Rotation`] in sync with `Transform`.
 ///
 /// ## See also
 ///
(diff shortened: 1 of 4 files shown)


Elabajaba:
Would this also add support for scaling colliders using bevy's `Transform`?

Jondolf:
Not yet, but I will soon add some `Scale` or `ColliderScale` component that syncs with `Transform` scale.

PR #99, opened 2023-07-27
Parallel Narrow Phase Collision Detection

TrustNoOneElse:
**Description:**
This draft pull request aims to enhance the performance of the collision detection system by addressing the bottleneck associated with the narrow phase collision detection using Parry. Currently, the narrow phase operates in a single-threaded manner and is tightly coupled with the penetration constraints, limiting the efficiency and the potential for parallel processing.

**Goals:**

- Separation of Narrow Phase into a Plugin: The major focus of this PR is to isolate the narrow phase collision detection into a dedicated "NarrowPhasePlugin." This plugin will handle the computation of contact points between colliding objects.
- Parallel Processing for Broad Phase Collision Pairs

I welcome every feedback and collaboration from everyone. Your insights or informations will be invaluable in identifying issues and refining the proposed changes since im just a beginner in this environment. But i want so badly to use bevy_xpbd in my project, so i dont accept yet a defeat for my project. Thanks for reading :) 

**Thougts/Notes from the maintainer:**
[see here](https://github.com/Jondolf/bevy_xpbd/pull/98#issuecomment-1652164202)

### DIFF
--- a/src/plugins/mod.rs
+++ b/src/plugins/mod.rs
@@ -24,6 +24,7 @@ pub mod broad_phase;
 #[cfg(feature = "debug-plugin")]
 pub mod debug;
 pub mod integrator;
+pub mod narrow_phase;
 pub mod prepare;
 pub mod setup;
 pub mod sleeping;
--- a/src/plugins/narrow_phase.rs
+++ b/src/plugins/narrow_phase.rs
@@ -1,0 +1,7 @@
+use bevy::prelude::*;
+
+pub struct NarrowPhasePlugin;
+
+impl Plugin for NarrowPhasePlugin {
+    fn build(&self, app: &mut App) {}
+}


Jondolf:
Hi, so I think I found the problem I was having in my previous attempts to separate the narrow phase from the solver.

Basically, this is the pseudocode for what we would like to have:

```rust
// Narrow phase
for pair in broad_collision_pairs {
    for contact in compute_contacts(pair) {
        contacts.push(contact);
    }
}

// Solver
for contact in contacts {
    solve_constraint(contact);
}
```

The problem is that this can actually miss collisions with the current implementation. The solver iterates through each contact and moves bodies to satisfy each constraint, but they can move into a penetrating state that wasn't detected by the narrow phase because they weren't colliding when it originally checked for collisions. This can cause bodies to accumulate high penetration depths without the collision being detected during the constraint solve, which makes them get high velocity the next frame and causes explosiveness.

One fix that I found is to give Parry's contact computation methods a prediction distance larger than 0. This way, it will also return collisions between bodies that aren't penetrating just yet, and we can check if they are actually penetrating in the constraints during each solve. This seems to resolve the issues I was having. The main "problem" is that it doesn't make sense to send collision events to users for bodies that aren't actually penetrating, so we might have to somehow filter out some of the events in the solver somehow.

I have a mostly working implementation of the `NarrowPhasePlugin` now, so I can make a separate PR for it. I think it would be good to split the `NarrowPhasePlugin` and parallelism into two separate PRs since they aren't necessarily related and they both have some different design considerations. For example, multithreading the narrow phase is likely to affect the constraint order, which would make the behavior undeterministic unless we sorted the contacts, and in some cases multithreading can actually be worse for performance due to the overhead, so it should be entirely optional.

Jondolf:
I made the PR: #100

TrustNoOneElse:
Thats pretty awesome thank you! I sadly didnt had time yet to do more experiments, but will continue with it soon! :) But then i will have something to be based on when i start, that will help for sure.

You are totally right, the multithreading should be a feature to be able to turn it on and off based on the usecase since it will not for everyone needed(Or maybe is it possible to detect if multithreading is needed and turn it on and off based on prevoius? Not sure how good this will work, but yeah i will experiment). Thank you for your insights about the design considerations. I will try to take care of once i begin. 

Just a simple thought for the prediction distance, shouldn't that be like based on the velocity of the object on how much i would predictable move in the next substep(maybe i get that wrong and thinking too easy on that)?

Jondolf:
> Just a simple thought for the prediction distance, shouldn't that be like based on the velocity of the object on how much i would predictable move in the next substep(maybe i get that wrong and thinking too easy on that)?

The narrow phase is inside the substepping loop and it runs *after* position integration (moving bodies based on velocity) so I don't think this is needed. However, the broad phase does do this for expanding AABBs since it's outside of the substepping loop and only runs once per physics frame.

The prediction distance needs to be there just because of the positional corrections that happen in the constraint solver loop, which runs directly after the narrow phase. At the start of the solver, two bodies might not be in contact, but other constraints can bring them into contact, and it's only detected if the prediction distance is large enough. I don't think the velocities of the bodies affect the prediction distance much in this case.

The prediction distance doesn't seem to impact performance much, so I think it's fine to choose a value that is "too large" in most cases. Currently I use 2.0 in 2D and 0.002 in 3D (different units), but we could probably even use 5.0 and 0.005 just to be safe.

TrustNoOneElse:
> The prediction distance doesn't seem to impact performance much, so I think it's fine to choose a value that is "too large" in most cases. Currently I use 2.0 in 2D and 0.002 in 3D (different units), but we could probably even use 5.0 and 0.005 just to be safe.
Sounds good! 

And thanks for the explaining, i thought a Substep would also process step by step the velocity^^

Jondolf:
Hi again, I had some problems with the performance of the separate narrow phase, so I optimized it a bit and also ended up implementing simple multithreading even though I originally said that it'd be done in another PR :P

Basically, `get_many` and `get_many_mut` seem to be quite expensive, so moving the narrow phase out of the solver actually made performance worse, because it needs a separate loop through the collision pairs instead of doing both in `penetration_constraints`.

This is still a problem to some extent, but I did some other optimization to the narrow phase, and now both the single-threaded and multi-threaded versions seem to be performing a bit better than the old approach. It's not a particularly significant difference though, and results are a bit contradictory since criterion benchmarks are indicating that performance is worse while Tracy and my own tests seem to be faster. I have more info in the PR itself under "Performance" (see #100).

I think I'll close this PR in favor of that one if that's fine

PR #105, opened 2023-08-02
feature: Add DistanceJoint and 2D and 3D examples.

shanecelis:
Hi,

This is a really cool project you've put together. I was using Rapier before this, and I really wanted a spring joint which isn't supported there. I hoped to find it here, but better than finding it I found that the XPBD means of expressing constraints and joints is so much easier to work with that I put one together myself that I hope you'll find useful. 

Taking note from other XPBD implementations, I called this a DistanceJoint. When it's compliance does not equal 0 though it very much behaves like a spring joint.

I added examples in 2D and 3D. Here are some gifs on them.

distance_joint_2d

distance_joint_3d

One problem I didn't tackle was if you look at the heading describing a SphericalJoint and a DistanceJoint, they're not terribly distinct. However, upon inspection SphericalJoint is doing more than DistanceJoint. I wonder if there's a clearer way to discriminate between them in their one line descriptions.

Thanks for the code, and thanks for turning me onto the the XPBD methodology. Happy hacking!

### DIFF
--- a/src/constraints/joints/distance.rs
+++ b/src/constraints/joints/distance.rs
@@ -1,0 +1,194 @@
+//! [`DistanceJoint`] component.
+
+use crate::prelude::*;
+use bevy::prelude::*;
+
+/// A distance joint prevents relative translation of the attached bodies away from its rest length while allowing rotation around all axes.
+///
+/// Distance joints can be useful for things like springs, muscles, and mass-spring networks.
+#[derive(Component, Clone, Copy, Debug, PartialEq)]
+pub struct DistanceJoint {
+    /// First entity constrained by the joint.
+    pub entity1: Entity,
+    /// Second entity constrained by the joint.
+    pub entity2: Entity,
+    /// Attachment point on the first body.
+    pub local_anchor1: Vector,
+    /// Attachment point on the second body.
+    pub local_anchor2: Vector,
+    /// The distance the attached bodies will be kept relative to each other.
+    pub rest_length: Scalar,
+    /// The extents of the allowed relative translation between the attached bodies.
+    pub length_limits: Option<DistanceLimit>,
+    /// Linear damping applied by the joint.
+    pub damping_linear: Scalar,
+    /// Angular damping applied by the joint.
+    pub damping_angular: Scalar,
+    /// Lagrange multiplier for the positional correction.
+    pub lagrange: Scalar,
+    /// The joint's compliance, the inverse of stiffness, has the unit meters / Newton.
+    pub compliance: Scalar,
+    /// The force exerted by the joint.
+    pub force: Vector,
+}
+
+impl XpbdConstraint<2> for DistanceJoint {
+    fn entities(&self) -> [Entity; 2] {
+        [self.entity1, self.entity2]
+    }
+
+    fn clear_lagrange_multipliers(&mut self) {
+        self.lagrange = 0.0;
+    }
+
+    fn solve(&mut self, bodies: [&mut RigidBodyQueryItem; 2], dt: Scalar) {
+        self.force = self.constrain_length(bodies, dt);
+    }
+}
+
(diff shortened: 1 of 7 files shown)


src/constraints/joints/distance.rs
```diff
@@ -0,0 +1,194 @@
+//! [`DistanceJoint`] component.
+
+use crate::prelude::*;
+use bevy::prelude::*;
+
+/// A distance joint prevents relative translation of the attached bodies away from its rest length while allowing rotation around all axes.
```
Jondolf, line 6: Just to make this a bit clearer and to make it more distinct from the spherical joint, I would have something like this:

```suggestion
/// A distance joint keeps the attached bodies at a certain distance from each other while while allowing rotation around all axes.
```

src/constraints/joints/distance.rs
```diff
@@ -0,0 +1,194 @@
...
+        self.lagrange = 0.0;
+    }
+
+    fn solve(&mut self, bodies: [&mut RigidBodyQueryItem; 2], dt: Scalar) {
+        self.force = self.constrain_length(bodies, dt);
+    }
+}
+
+impl Joint for DistanceJoint {
+    fn new(entity1: Entity, entity2: Entity) -> Self {
```
Jondolf, line 50: I noticed that `Joint` must implement `new` like this, but it would probably be nice to be able to specify the rest length here directly since it's the main point of the distance joint.

This is fine for now, but I can make a follow-up PR that removes `new` from `Joint` to allow joints to have custom constructor methods.

Jondolf:
Thanks, looks great! I'm glad that the constraints are understandable enough that people can make new constraints like this.

The main difference between this distance constraint and the spherical joint is that the spherical joint doesn't allow any relative translation, while the distance constraint allows 1 or 2 (2D/3D) translational DOFs. The spherical joint also has optional swing and twist limits. I added a comment to clarify the `DistanceJoint` description a bit.

src/constraints/joints/distance.rs
```diff
@@ -0,0 +1,194 @@
...
+        self.lagrange += delta_lagrange;
+
+        // Apply positional correction (method from PositionConstraint)
+        self.apply_positional_correction(body1, body2, delta_lagrange, n, world_r1, world_r2);
+
+        // Return constraint force
+        self.compute_force(self.lagrange, n, dt)
+    }
+
+    /// Sets the translational limits along the joint's free axis.
```
Jondolf, line 177:
```suggestion
    /// Sets the minimum and maximum distances between the attached bodies.
```

src/constraints/joints/distance.rs
shanecelis: I think a distance joint with zero rest length is valid though, so I wouldn't be eager to remove new (). It's nice to have a unified constructor for all joints.

shanecelis:
Thanks for the kind words and the explanation for the difference between distance joint and spherical joint. 

Honestly after I wrote the code and started writing the documentation, I had a jolt when I saw spherical joint's description and I wondered if I had just reimplemented it for no benefit (other than my own challenge). Glad that wasn't the case.

Look forward to putting this to use. Thanks for the speedy reply. I'm impressed with the pace of work I've seen on this project.

src/constraints/joints/distance.rs
Jondolf: It's valid, but most people probably want to specify a non-zero rest length for distance joints, otherwise you could use some other joint like a spherical joint. But I'll leave it as it is for now, I agree that it's nice to have a unified API.

shanecelis:
Yes. Absolutely. -ShaneOn Aug 2, 2023, at 9:20 AM, Joona Aalto ***@***.***> wrote:﻿
Thanks :)
Could you resolve the lint warnings so that I can get this merged? I think it's mostly whitespace stuff.

PR #107, opened 2023-08-05
Added text that explained how to use the example 'chain_3d'

Aztro-dev:
This change adds a ui that explains how to move around the chain in the `chain_3d`  example in `bevy_xpbd_3d`
The text says `"Use Arrow Keys to Move The Chain"`

### DIFF
--- a/crates/bevy_xpbd_3d/examples/chain_3d.rs
+++ b/crates/bevy_xpbd_3d/examples/chain_3d.rs
@@ -126,8 +126,8 @@ fn ui(mut commands: Commands) {
                 )
                 .with_style(Style {
                     margin: UiRect {
-                        left: Val::Percent(0.5),
-                        top: Val::Percent(2.0),
+                        left: Val::Px(5.0),
+                        top: Val::Px(30.0),
                         ..default()
                     },
                     ..default()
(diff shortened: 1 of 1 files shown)


crates/bevy_xpbd_3d/examples/chain_3d.rs
```diff
@@ -95,3 +95,47 @@ fn movement(
...
+                    TextStyle {
+                        font_size: 20.0,
+                        color: Color::WHITE,
+                        ..default()
+                    },
+                )
+                .with_style(Style {
+                    margin: UiRect {
+                        left: Val::Percent(0.5),
+                        top: Val::Percent(2.0),
```
Jondolf, line 130: This makes the label overlap the FPS counter when the window width is small:

Text overlap

I haven't used Bevy UI much so I'm not sure what the best practises are, but using pixels for the offsets makes the label consistent with the FPS counter:

```suggestion
                        left: Val::Px(5.0),
                        top: Val::Px(30.0),
```

Jondolf:
Thanks, we should probably eventually add help labels like this for all relevant examples. I left a comment about text overlap, I can merge this once that's resolved

Aztro-dev:
I accepted the change, this should be ready to merge!

PR #128, opened 2023-08-19
Correct time dilation for FixedUpdate schedule.

### LINKED ISSUE #124: Incorrect time dilation for custom fixed schedules.
Configurations with their own fixed schedule passed to `PhysicsPlugins::new()` have incorrect time dilation.
In my case I pass Bevy's `FixedUpdate` schedule that I have configurated to run at `1./16.`. This causes physics sim to run in slow-motion. Setting `PhysicsTimestep::Fixed` in addition does not solve the time dilation problem.
I pass my own `FixedUpdate` to keep my [schedules in sync](https://docs.rs/bevy_xpbd_3d/latest/bevy_xpbd_3d/plugins/struct.PhysicsPlugins.html#custom-schedule).

starwolfy:
Fixes #124 .

### DIFF
--- a/src/plugins/setup.rs
+++ b/src/plugins/setup.rs
@@ -116,7 +116,6 @@ impl Plugin for PhysicsSetupPlugin {
         );
 
         // Check and store if schedule is configured to run in FixedUpdate.
-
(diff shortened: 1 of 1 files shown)


Jondolf:
Thanks, it looks like everything works as expected :)

PR #130, opened 2023-08-21
add Rotation::from_sin_cos constructor

RJ:
Being able to construct Rotation from sin and cos allows for full-precision serialization and deserialization, without having to (eg) go via a single scalar radians value.


src/components/rotation.rs
```diff
@@ -101,6 +101,11 @@ impl Rotation {
         }
     }
 
+    /// Creates a [`Rotation`] from sine and cosine in radians
```
Jondolf, line 104: Just to be slightly more descriptive and consistent:
```suggestion
    /// Creates a [`Rotation`] from the sine and cosine of an angle in radians.
```

RJ:
committed your suggested change, then force pushed after squashing into one commit

PR #133, opened 2023-08-26
Avoid allocate on spatial queries. ray_hits_callback, shape_hits_callback, intersections_callback

bnyu:
When users only want do callback on spatial queries, avoid unnecessary allocate. (ray_hits,shape_hits,point_intersections still return results)


Jondolf:
Ah, I think lint is failing because Rust 1.72.0 added support for `let else` formatting

bnyu:
Ah, I haven't upgraded to 1.72.0 yet. I just run cargo fmt at 1.71.1 and force pushed.

Jondolf:
I fixed the formatting on main now

Jondolf:
The changes look good to me, I'll rebase and merge

PR #134, opened 2023-08-28
Fix 2d `position_to_transform` changing child `Transform` z-component

### LINKED ISSUE #125: 2D child Z grows without bounds
I made an entity with a position that has a child which also has a position. Both of them have a nonzero z component for draw ordering. That caused the z component of the transform to go off like crazy.

I think this is caused by `position_to_transform` taking the z coordinate of the child's original transform and treating it as a global transform. 
```
                let new_transform = GlobalTransform::from(
                    Transform::from_translation(pos.as_f32().extend(transform.translation.z))
                        .with_rotation(Quaternion::from(*rot).as_f32()),
                )
                .reparented_to(&GlobalTransform::from(parent_transform));
```

I changed this to 
```
                let new_transform = GlobalTransform::from(
                    Transform::from_translation(pos.as_f32().extend(transform.translation.z + parent_transform.translation.z))
                        .with_rotation(Quaternion::from(*rot).as_f32()),
                )
                .reparented_to(&GlobalTransform::from(parent_transform));
```
but I don't know if that still works if the parent is scaled. Maybe it would be best to just set it after doing the `GlobalTransform::from`?

ShaddowSpy:
Fixes #125 
child's `Transform.translation.z` was previously added twice, in `extend()` and `reparented_to()`
This change removes the addition that was in `extend()`.

### DIFF
--- a/src/plugins/sync.rs
+++ b/src/plugins/sync.rs
@@ -245,7 +245,7 @@ fn position_to_transform(
                 // The new local transform of the child body,
                 // computed from the its global transform and its parents global transform
                 let new_transform = GlobalTransform::from(
-                    Transform::from_translation(pos.as_f32().extend(transform.translation.z))
+                    Transform::from_translation(pos.as_f32().extend(0.0))
                         .with_rotation(Quaternion::from(*rot).as_f32()),
                 )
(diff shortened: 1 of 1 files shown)


Jondolf:
Thanks!

PR #135, opened 2023-08-29
async_colliders example in 3D

dt665m:
This PR adds an example that could help debug some transform inconsistencies when creating colliders asynchronously.


Jondolf:
Closing since the issue was fixed and async colliders were added in #190.

PR #139, opened 2023-08-30
Implement PartialEq for Collider

RJ:
All the `bevy_xpbd` components I'm using implement `PartialEq` except for `Collider`. 

Having `PartialEq` is really useful for various netcode bits that check what's changed in generic systems.

`Collider` is a wrapper over parry's `SharedShape`, and not all parry `Shape`s implement `PartialEq`

For those that do, i'm using it. For the rest, i'm comparing what is necessary manually – eg, border radius for rounded versions of shapes.

I took the big match on shape types from the implementation of `fmt::Debug` for `Collider`, so `PartialEq` will work for all the same shapes.

### DIFF
--- a/src/components/collider.rs
+++ b/src/components/collider.rs
@@ -126,6 +126,155 @@ impl fmt::Debug for Collider {
     }
 }
 
+/// Compares two [`parry::SharedShape`], for Collider's PartialEq implementation.
+///
+/// Proxy for `PartialEq` for those `parry::Shape` that implement PartialEq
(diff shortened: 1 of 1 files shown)


Jondolf:
Thanks! Ideally, this would be implemented for Parry's `SharedShape` directly, but this works as a workaround for now. We could also make a PR to add this to Parry though.

RJ:
https://github.com/dimforge/parry/pull/162

PR #145, opened 2023-09-02
Fix math cross-platform determinism

Jondolf:
Enables the `libm` feature for glam to support cross-platform deterministic math when `enhanced-determinism` is enabled.

Without the `libm` feature, our determinism test produces different snapshots across Windows, Ubuntu and MacOS. This can be seen as the CI failing in the second commit. Enabling `libm` makes the snapshots consistent across operating systems.

Our test is quite straightforward however, as it just spawns a stack of cubes that falls to the ground. It doesn't test if joints, spatial queries, sleeping etc. are deterministic. We should make the test more complex to cover these different cases and verify proper determinism.

Note that even if this does fix determinism for bevy_xpbd, it's still up to the users to make sure their own code is deterministic. We should add proper documentation and tips on determinism once it's verified to work for the engine itself.


johanhelsing (approved):
I think there are still unsolved determinism issues, but this is a step in the right direction :)

Issue #110, opened 2023-08-07
Panic when removing colliders


LeshaInc:
Sometimes when removing colliders I get panic at `src/plugins/spatial_query/pipeline.rs:135:67`:

```rust
fn update_incremental_internal(...) {
    // ...

    if refit_and_balance {
        let _ = self.qbvh.refit(0.0, &mut self.workspace, |entity_index| {
            // Construct entity ID
            let generation = self.entity_generations.get(entity_index).map_or(0, |i| *i);
            let entity = utils::entity_from_index_and_gen(*entity_index, generation);
            // Compute and return AABB
            let (iso, shape, _) = self.colliders.get(&entity).unwrap(); // <-- here
            let aabb = shape.get_shape().compute_aabb(iso);
            aabb
        });
        self.qbvh.rebalance(0.0, &mut self.workspace);
    }

    // ...
}
```

Happens on 6ec8bc4

PR #152, opened 2023-09-10
Avoid incremental updates to Qbvh

NiseVoid:
The incremental updates on Qbvh don't work as intended, an issue about this has been open on rapier/parry for a long time but the issues are yet to be solved. For now it's faster and more stable to rebuild the Qbvh entirely ever frame.

### DIFF
--- a/src/plugins/spatial_query/mod.rs
+++ b/src/plugins/spatial_query/mod.rs
@@ -159,7 +159,7 @@ pub use shape_caster::*;
 pub use system_param::*;
 
 use crate::prelude::*;
-use bevy::{prelude::*, utils::HashSet};
+use bevy::prelude::*;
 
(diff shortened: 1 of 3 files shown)


Jondolf:
Thanks, this seems to fix a lot of the issues regarding crashes and increasing memory usage without noticably degrading performance. This probably also closes #110 since the `unwrap` call no longer exists.

I would expect that incremental updates *should* be more efficient when done properly, so we should revisit this later, at least once dimforge/parry#146 and dimforge/rapier#450 are resolved.

I noticed that `RemovedColliders` doesn't seem to be used anywhere anymore and it's a private resource. Could you maybe remove it and the systems associated with it? It'll probably need to be added back at some point, but right now it has no purpose.

Issue #150, opened 2023-09-09
Collision Filtering and Modification


datael:
Future features in the readme lists Collision Filtering as something not yet implemented and I would like to get the ball rolling if possible.

### Topics discussed in the xpbd help channel on Discord

Some topics that were discussed on Discord included:
1. Global physics hooks that can iterate over all collisions
2. Per-entity filter components
    * Filtering collisions with entities by EntityId
    * Filtering collisions by predicate
3. Being able to modify the individual contacts
4. The possibility of being able to use one-shot systems in the future
5. The possibility of being able to use entity relations to allow more efficient filtering

### Global Physics hooks
Adding a schedule after the `SubstepSet::NarrowPhase`, and exposing a way to filter collisions from the `Collisions` resource gives a very open-ended and straightforward way to allow for modifying collisions.
I've got a working example [on this fork here](https://github.com/datael/bevy_xpbd/commits/collision_post_processing).

I'm in two minds about whether it's worth the schedule to have built-in sets (e.g. `First`, `Filter`, `Modify`, `Last`), or whether that complexity should be left up to the user. I can see a potential want to pre-filter collisions with faster filters before any slower ones execute, for example.

### Per-entity filter components
@Jondolf suggested something in the following form:
```rust
ContactFilter::new()
        .with_excluded_entities([...])
        .with_predicate(|mut contact: ContactData| {
            // Modify contact (optional)
            contact.penetration = 0.1;
            // Return the contact or use None to ignore contact
            Some(contact)
        }),
```

It was also suggested that some form of combination rule may be needed to handle when two entities with conflicting filter predicates are present.

### Modifying individual contacts
Being able to modify the individual contacts is desirable. Therefore, being able to return `Some(contact)` from a post-processing step is more useful than only being able to remove or keep collisions.

### One-shot systems
@Jondolf showed the following code on Discord as something that it would be nice to be able to do. There was discussion about how this may not be possible with one-shot systems, but using `run_system_with`* may allow it as long as the overhead was acceptable.

\* `run_system_with` was added to bevy with [this merged but unreleased-at-the-time-of-writing pull request](https://github.com/bevyengine/bevy/pull/9366).

```rust
fn setup(mut commands: Commands, contact_hook_pipeline: ContactHookPipeline) {
    commands.spawn((
        ContactHook::new(
            &contact_hook_pipeline,
            |In(mut contact): In<ContactData>, query: Query<&MyBundle>| {
                // Modify contact, query for entities, do whatever
[...]

PR #155, opened 2023-09-17
Collision post processing

datael:
## Objective
Create a stage to allow filtering and changing collisions from user-defined systems.

This is the simplest part of #150: the Global Physics hooks via user systems only.

## Solution
- Adds `SubstepSet::PostProcessCollisions`, which is directly after `SubstepSet::NarrowPhase`
- Adds `PostProcessCollisionsSchedule`, which is run in `SubstepSet::PostProcessCollisions`
- Exposes `retain` from `Collisions` to allow direct modification of the set of collisions

Also includes a simple one-way-platform example.

### DIFF
--- a/crates/bevy_xpbd_2d/examples/one_way_platform_2d.rs
+++ b/crates/bevy_xpbd_2d/examples/one_way_platform_2d.rs
@@ -1,0 +1,327 @@
+use bevy::{prelude::*, sprite::MaterialMesh2dBundle, utils::HashSet};
+use bevy_xpbd_2d::{math::*, prelude::*, PostProcessCollisionsSchedule};
+use examples_common_2d::XpbdExamplePlugin;
+
+fn main() {
+    App::new()
+        .add_plugins((DefaultPlugins, XpbdExamplePlugin))
+        .insert_resource(ClearColor(Color::rgb(0.05, 0.05, 0.1)))
+        .insert_resource(SubstepCount(6))
+        .insert_resource(Gravity(Vector::NEG_Y * 1000.0))
+        .add_systems(Startup, setup)
+        .add_systems(Update, (movement, pass_through_one_way_platform))
+        .add_systems(PostProcessCollisionsSchedule, one_way_platform)
+        .run();
+}
+
+#[derive(Component)]
+struct Actor;
+
+#[derive(Clone, Eq, PartialEq, Debug, Default, Component)]
+pub struct OneWayPlatform(HashSet<Entity>);
+
+#[derive(Copy, Clone, Eq, PartialEq, Debug, Default, Component, Reflect)]
+pub enum PassThroughOneWayPlatform {
+    #[default]
+    /// Passes through a `OneWayPlatform` if the contact normal is in line with the platform's local-space up vector
+    ByNormal,
+    /// Always passes through a `OneWayPlatform`, temporarily set this to allow an actor to jump down through a platform
+    Always,
+    /// Never passes through a `OneWayPlatform`
+    Never,
+}
+
+fn setup(
+    mut commands: Commands,
+    mut materials: ResMut<Assets<ColorMaterial>>,
+    mut meshes: ResMut<Assets<Mesh>>,
+) {
+    commands.spawn(Camera2dBundle::default());
+
+    // For borders
+    let square_sprite = Sprite {
+        color: Color::rgb(0.7, 0.7, 0.8),
+        custom_size: Some(Vec2::splat(50.0)),
+        ..default()
+    };
+
+    // Ceiling
+    commands.spawn((
+        SpriteBundle {
+            sprite: square_sprite.clone(),
+            transform: Transform::from_scale(Vec3::new(20.0, 1.0, 1.0)),
+            ..default()
+        },
+        RigidBody::Static,
+        Position(Vector::Y * 50.0 * 6.0),
+        Collider::cuboid(50.0 * 20.0, 50.0),
+    ));
+    // Floor
+    commands.spawn((
+        SpriteBundle {
+            sprite: square_sprite.clone(),
+            transform: Transform::from_scale(Vec3::new(20.0, 1.0, 1.0)),
+            ..default()
+        },
+        RigidBody::Static,
+        Position(Vector::NEG_Y * 50.0 * 6.0),
+        Collider::cuboid(50.0 * 20.0, 50.0),
+    ));
+    // Left wall
+    commands.spawn((
+        SpriteBundle {
+            sprite: square_sprite.clone(),
+            transform: Transform::from_scale(Vec3::new(1.0, 11.0, 1.0)),
+            ..default()
+        },
+        RigidBody::Static,
+        Position(Vector::NEG_X * 50.0 * 9.5),
+        Collider::cuboid(50.0, 50.0 * 11.0),
+    ));
+    // Right wall
+    commands.spawn((
+        SpriteBundle {
+            sprite: square_sprite,
+            transform: Transform::from_scale(Vec3::new(1.0, 11.0, 1.0)),
+            ..default()
+        },
+        RigidBody::Static,
+        Position(Vector::X * 50.0 * 9.5),
+        Collider::cuboid(50.0, 50.0 * 11.0),
+    ));
+
+    // For one-way platforms
+    let one_way_sprite = Sprite {
+        color: Color::rgba(0.7, 0.7, 0.8, 0.25),
+        custom_size: Some(Vec2::splat(50.0)),
+        ..default()
+    };
+
+    // Spawn some one way platforms
+    for y in -2..=2 {
+        commands.spawn((
+            SpriteBundle {
+                sprite: one_way_sprite.clone(),
+                transform: Transform::from_scale(Vec3::new(10.0, 0.5, 1.0)),
+                ..default()
+            },
+            RigidBody::Static,
+            Position(Vector::Y * 15.0 * 6.0 * y as f32),
+            Collider::cuboid(25.0 * 20.0, 25.0),
+            OneWayPlatform::default(),
+        ));
+    }
+
+    // Spawn an actor for the user to control
+    let actor_size = Vec2::new(20.0, 20.0);
+    let actor_mesh = MaterialMesh2dBundle {
+        mesh: meshes.add(shape::Quad::new(actor_size).into()).into(),
+        material: materials.add(ColorMaterial::from(Color::rgb(0.2, 0.7, 0.9))),
+        ..default()
+    };
+
+    commands.spawn((
+        actor_mesh.clone(),
+        RigidBody::Dynamic,
+        LockedAxes::ROTATION_LOCKED,
+        Position(Vector::ZERO),
+        Collider::cuboid(actor_size.x, actor_size.y),
+        Actor,
+        PassThroughOneWayPlatform::ByNormal,
+    ));
+}
+
+fn movement(
+    keyboard_input: Res<Input<KeyCode>>,
+    mut actors: Query<&mut LinearVelocity, With<Actor>>,
+) {
+    for mut linear_velocity in &mut actors {
+        if keyboard_input.pressed(KeyCode::Left) {
+            linear_velocity.x -= 10.0;
+        }
+        if keyboard_input.pressed(KeyCode::Right) {
+            linear_velocity.x += 10.0;
+        }
+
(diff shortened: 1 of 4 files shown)


datael:
In https://github.com/Jondolf/bevy_xpbd/pull/156 @Jondolf said they were planning on exposing normal2. After normal2 is exposed, the one way platform example can be simplified, so returning to draft state pending https://github.com/Jondolf/bevy_xpbd/pull/156 getting merged.

Jondolf:
#156 is now merged! It exposes `normal2` for contacts as I mentioned. I can review this soon :)

datael:
Ok, that sample is now updated to use normal2 where appropriate.
Thank you for adding that. Now that the normal doesn't need to be rotated, the one way platform example is a lot less math-heavy!

src/lib.rs
```diff
@@ -464,6 +464,12 @@ pub struct PhysicsSchedule;
 #[derive(Debug, Hash, PartialEq, Eq, Clone, ScheduleLabel)]
 pub struct SubstepSchedule;
 
+/// The schedule that runs in [`SubstepSet::PostProcessCollisions`].
+///
+/// Empty by default.
+#[derive(Debug, Hash, PartialEq, Eq, Clone, ScheduleLabel)]
+pub struct PostProcessCollisionsSchedule;
```
Jondolf, line 471: `PostProcessCollisionsSchedule` feels a bit verbose and hard to read due to the length in my opinion.

For clarity and to avoid naming conflicts, we currently have e.g. `PhysicsSchedule` and `SubstepSchedule`, but maybe in this case it could be just `PostProcessCollisions` (without the `Schedule`)? Bevy doesn't use `Schedule` in its own names like `Update` and even `Main`, so I think it'd be better to leave it out for brevity.

```suggestion
pub struct PostProcessCollisions;
```

Alternatively, this could be `CollisionHooks` or `CollisionHookSchedule`, but I'm not sure if these types of systems count as hooks necessarily, so I do like the `PostProcess` name since it perhaps makes less assumptions.

src/lib.rs
```diff
@@ -548,6 +558,14 @@ pub enum SubstepSet {
     ///
     /// See [`NarrowPhasePlugin`].
     NarrowPhase,
+    /// Responsible for running [`PostProcessCollisionsSchedule`] to allow user-defined systems
+    /// to post-process collisions, such as filtering.
```
Jondolf, line 562: In "post-process collisions, such as filtering", the "such as" seems to be referring to the wrong thing (collisions instead of post-processing).

```suggestion
    /// Responsible for running the [`PostProcessCollisions`] schedule to allow user-defined systems
    /// to filter and modify collisions.
```

Here I assume that the schedule name was changed; if we land on something else, this will need to be changed of course

src/plugins/setup.rs
```diff
@@ -181,6 +184,27 @@ impl Plugin for PhysicsSetupPlugin {
             PhysicsSchedule,
             run_substep_schedule.in_set(PhysicsStepSet::Substeps),
         );
+
+        // Create post-process collisions schedule, the schedule that should be used to
+        // run collision post-processing systems, such as filtering.
```
Jondolf, line 189:
```suggestion
        // Create collision post-processing schedule for user-defined systems
        // that filter and modify collisions.
```

crates/bevy_xpbd_2d/examples/one_way_platform_2d.rs
```diff
@@ -0,0 +1,305 @@
...
+    // Spawn some one way platforms
+    for y in -2..=2 {
+        commands.spawn((
+            SpriteBundle {
+                sprite: one_way_sprite.clone(),
+                transform: Transform::from_scale(Vec3::new(10.0, 0.5, 1.0)),
+                ..default()
+            },
+            RigidBody::Static,
+            Position(Vector::Y * 15.0 * 6.0 * y as Scalar),
```
Jondolf, line 109: The distance between the ground and the first platform seems to be 15px larger than the distance between the platforms. We should probably either move the ground up or the platforms down.

```suggestion
            Position(Vector::Y * 15.0 * 6.0 * y as Scalar - Vector::Y * 15.0),
```

crates/bevy_xpbd_2d/examples/one_way_platform_2d.rs
```diff
@@ -0,0 +1,305 @@
...
+        }
+        if keyboard_input.pressed(KeyCode::Right) {
+            linear_velocity.x += 10.0;
+        }
+
+        if linear_velocity.y.abs() < 0.1 // Assume "mostly stopped" to mean "grounded"
+            && !keyboard_input.pressed(KeyCode::Down)
+            && keyboard_input.just_pressed(KeyCode::Space)
+        {
+            linear_velocity.y = 500.0;
```
Jondolf, line 150: The jump height feels just a bit too large in my opinion; maybe reduce this to 450.0 or increase the gravity a bit?
```suggestion
            linear_velocity.y = 450.0;
```

crates/bevy_xpbd_2d/examples/one_way_platform_2d.rs
```diff
@@ -0,0 +1,305 @@
...
+        mesh: meshes.add(shape::Quad::new(actor_size).into()).into(),
+        material: materials.add(ColorMaterial::from(Color::rgb(0.2, 0.7, 0.9))),
+        ..default()
+    };
+
+    commands.spawn((
+        actor_mesh.clone(),
+        RigidBody::Dynamic,
+        LockedAxes::ROTATION_LOCKED,
+        Position(Vector::ZERO),
```
Jondolf, line 127: `Position` is already at the origin, so this is unnecessary. I would specify `Restitution::ZERO` for the player though since it probably shouldn't bounce unless the material is specifically bouncy

```suggestion
        Restitution::ZERO.with_combine_rule(CoefficientCombine::Min),
```

crates/bevy_xpbd_2d/examples/one_way_platform_2d.rs
```diff
@@ -0,0 +1,305 @@
...
+) {
+    for mut linear_velocity in &mut actors {
+        if keyboard_input.pressed(KeyCode::Left) {
+            linear_velocity.x -= 10.0;
+        }
+        if keyboard_input.pressed(KeyCode::Right) {
+            linear_velocity.x += 10.0;
+        }
+
+        if linear_velocity.y.abs() < 0.1 // Assume "mostly stopped" to mean "grounded"
```
Jondolf, line 146: I'm fine with this for simplicity, but this does allow you to infinitely jump in the air if you time your jumps well or just spam jump fast enough. Maybe mention better alternatives?
```suggestion
        // Assume "mostly stopped" to mean "grounded".
        // You should use ray casting, shape casting or sensor colliders
        // for more robust ground detection.
        if linear_velocity.y.abs() < 0.1
```

crates/bevy_xpbd_2d/examples/one_way_platform_2d.rs
```diff
@@ -0,0 +1,305 @@
...
+    mut one_way_platforms_query: Query<&mut OneWayPlatform>,
+    other_colliders_query: Query<
+        Option<&PassThroughOneWayPlatform>,
+        (With<Collider>, Without<OneWayPlatform>), // NOTE: This precludes OneWayPlatform passing through a OneWayPlatform
+    >,
+    mut collisions: ResMut<Collisions>,
+) {
+    // This assumes that Collisions contains empty entries for entities
+    // that were once colliding but no longer are.
+
```
Jondolf, line 229: Is this line break intended? Looks a bit out of place to me, but maybe it's just preference

```suggestion
```

crates/bevy_xpbd_2d/examples/one_way_platform_2d.rs
```diff
@@ -0,0 +1,305 @@
+use bevy::{prelude::*, sprite::MaterialMesh2dBundle, utils::HashSet};
```
Jondolf, line 1:
```suggestion
//! A 2D platformer example with one-way platforms to demonstrate
//! filtering collisions with systems in the `PostProcessCollisions` schedule.
//!
//! Move with arrow keys, jump with Space and descend through
//! platforms by pressing Space while holding the down arrow.

use bevy::{prelude::*, sprite::MaterialMesh2dBundle, utils::HashSet};
```

Jondolf (commented):
I left quite a few comments, mostly just additional code comments or tweaks to language/naming and some of the values in the example.

The one-way platform example is *excellent*, and the system could probably be just dropped into most projects with very little changes. I'm happy to merge this once we've sorted out the things I mentioned in my review :)

src/lib.rs
datael: Agree with you. I was following suit with the other schedules, but agree that it's too verbose as `PostProcessCollisionsSchedule`.

One thing I am slightly concerned about is that in #150 we are discussing filtering during the broad phase as well.
I wonder if we would end up wanting a similar name in there, too?

E.g. maybe `PostProcessNarrowPhase` is more appropriate when we consider potentially wanting `PostProcessBroadPhase` in the future too?

src/lib.rs
datael: Agree with your changes here.
I believe your assumption is correct. At the time I think I was trying to match your original wording for the comments, but I probably took it too far.

crates/bevy_xpbd_2d/examples/one_way_platform_2d.rs
datael: This was intentional, since it applies to the whole function but is less important than the full explanation, but if the extra empty line is distracting (it must be for you to have noticed it) then let's remove it.

crates/bevy_xpbd_2d/examples/one_way_platform_2d.rs
datael: I was going to disagree with you on this because I preferred the way it looked wrt the borders at 15 spacing, but after changing jump strength to 450 (which I did agree with) we can no longer reach the bottom platform from the ground(!)

I'm not 100% sure myself, but this does _appear_ to be the correct height so that the spacing is even, so let's go with this.

src/lib.rs
Jondolf: For the broad phase filtering, I meant just checking if `excluded_entities.contains(&entity)` in the existing systems just like how we filter by collision layers. I don't think we necessarily need any separate filtering step in the broad phase.

I'm also not entirely sure if the separate component I suggested for excluding entities is even necessary, since you could just use collision layers (or in the future maybe custom predicates) to have the same effect.

In my opinion, the only other place where a filtering step might make sense is the solver, i.e. penetration constraints. They contain the actual contact forces, and it could be useful to e.g. skip contacts when some force threshold is exceeded. However I think this would require actual hooks/callbacks, because the constraints are currently applied as soon as they're created, so a separate step wouldn't really work.

So I would just name it as `PostProcessCollisions` for now, since I find it unlikely that we add other separate filtering steps anytime soon, and if we do, we can just rename it again.

datael: Ok that makes sense 👍 Thanks for such a quick reply!

src/plugins/setup.rs
```diff
@@ -300,3 +321,9 @@ fn run_substep_schedule(world: &mut World) {
         world.run_schedule(SubstepSchedule);
     }
 }
+
+/// Runs the [`PostProcessCollisions`] schedule.
+fn run_post_process_collisions_schedule(world: &mut World) {
+    debug!("running PostProcessCollisions");
+    world.run_schedule(PostProcessCollisions);
+}
```
datael, line 329: This is bikeshedding now, but I'm in two minds about whether this function should be 
* `run_post_process_collisions_schedule` or
* `run_post_process_collisions`

Similarly, for the log, since we've dropped `Schedule` from the name, should it be
* `"running PostProcessCollisions"` or
* `"running PostProcessCollisions schedule"`

Jondolf: I might slightly prefer `run_post_process_collisions` and `"running PostProcessCollisions"` for brevity, but I have no strong preference either way. It's not very user-facing so it doesn't really matter

Jondolf: Actually no, maybe `run_post_process_collisions_schedule` to make it clearer that that system doesn't actually do the post-processing

Jondolf: Yep, looks good 👍

Jondolf (approved):
Looks good now, thanks!

PR #167, opened 2023-10-05
Try to fix Ubuntu CI test suite

Jondolf:
I'm trying to fix CI errors, will attempt different stuff until it works


kav:
have you tried setting jobs=1 on ubuntu? I kind of wonder if the cpu count is either wrong or not a great metric for job count on the linux runners. I had some success with ubuntu ci working with act https://github.com/nektos/act when I set jobs to 1 but YMMV

Jondolf:
#729 added Linux builds back :)

PR #169, opened 2023-10-07
Add contact queries

### LINKED ISSUE #168: Add `contacts` method to SpatialQuery
Hi ! I'm trying to make my local fork of `bevy_mod_wanderlust` work with `bevy_xpbd` and one of the things that's tripping me up is that I can't find any direct parallel for `bevy_rapier3d::parry::query::contact`. `Bevy_mod_wanderlust` does some interesting iterative position adjusting in the case they're penetrating the ground already and relies on `contact` to know which direction to back out from:

```rs
if let Ok(Some(contact)) = bevy_rapier3d::parry::query::contact(
    &cast_iso,
    &*shape_caster.raw,
    &ground_iso,
    &*ground_collider.raw,
    0.0,
) {
    // Adjust position so we're no longer contacting
}
```

`bevy_xpbd` has a similar `SpatialQuery::shape_intersections` but it only returns an Entity. I think I could align the entities with the Contacts in the physics world or something but I think they'd be out of date with those from the shape_intersections if you we're to run it multiple times in one frame meanwhile adjusting positions right? That's just my assumption, I'm not sure.

### Edit (I'm dumb)

`shape_intersections` I think still should return Contacts but that function isn't what `bevy_rapier3d::parry::query::contact` does. `shape_intersections` returns all entities colliding with that one specific shape. Instead, I think it'd be better to ask for a function to be added to SpatialQuery that returns the contact points between two colliders given their isometries? I might be able to PR this if you'd prefer.

Anyways, very much enjoying the library :D

Jondolf:
Closes #168.

## Objective

It can often be useful to compute contacts and other geometric information for specific pairs of colliders instead of having to query for a ton of entities. Just being able to compute a contact pair or a time of impact for shapes A and B at specific positions can be very powerful.

This functionality is provided by Parry's methods like `contact` and `time_of_impact`, but it's not exposed in our APIs and requires dealing with Parry's own types. We should have our own functions that hide the unnecessary implementation details.

## Solution

This PR creates a `contact_query` module in `narrow_phase` that contains equivalents for Parry's geometric queries like `contact`, `contact_manifolds`, `time_of_impact` and so on.

`ContactManifold` also no longer contains the colliding entities, because it's not needed internally, and you shouldn't need to provide any entities for `contact_query::contact_manifolds`.

## Implemented queries

- [x] `contact`
- [x] `contact_manifolds`
- [x] `closest_points`
- [x] `distance`
- [x] `intersection_test`
- [x] `time_of_impact`
- [ ] (`nonlinear_time_of_impact`)

`nonlinear_time_of_impact` is probably niche enough that we can add it if/when there is a need for it.

## Thoughts

This PR has these contact queries under a separate module as standalone functions, but I can see why it could also make sense to have all geometric queries under one place, aka `SpatialQuery`. However, I feel like `SpatialQuery` would add unnecessary restrictions, since these contact queries don't have any external state and don't need access to the ECS world, and it's good to be able to use contact queries even when `SpatialQueryPlugin` is disabled. We can also make the docs a bit nicer this way since the API used for "real spatial queries" and these contact queries is inherently a bit different. This is why I prefer having them under a separate module.


Jondolf:
I have now added all of the contact queries that we will probably need for now, along with proper documentation and code examples.

(note: CI is failing because of unrelated storage issues, so we can probably merge this since the tests are passing locally)

stargazing-dino (approved):
Really good looking docs and examples !

PR #173, opened 2023-10-10
Skip re-writing the same values to components, to satisfy bevy's change detection.

RJ:
At the moment, a dynamic body (that isn't sleeping) with zero Linear and Angular velocity will still get its Position, Rotation, LinearVelocity, and AngularVelocity updated every tick (re-set to the same values).

This shows up if you use a `Query<Entity, Changed<Position>>` query filter, for example.

These commits guard against updating the components unless the value will actually change.

I noticed this while debugging some netcode that serializes and sends changed component values.

### DIFF
--- a/src/plugins/solver.rs
+++ b/src/plugins/solver.rs
@@ -220,7 +220,11 @@ fn update_lin_vel(
 
         if rb.is_dynamic() {
             // v = (x - x_prev) / h
-            lin_vel.0 = (pos.0 - prev_pos.0 + translation.0) / sub_dt.0;
+            let new_lin_vel = (pos.0 - prev_pos.0 + translation.0) / sub_dt.0;
+            // avoid triggering bevy's change detection unnecessarily
+            if new_lin_vel != lin_vel.0 {
+                lin_vel.0 = new_lin_vel;
+            }
         }
     }
 }
@@ -249,7 +253,11 @@ fn update_ang_vel(
         pre_solve_ang_vel.0 = ang_vel.0;
 
         if rb.is_dynamic() {
-            ang_vel.0 = (rot.mul(prev_rot.inverse())).as_radians() / sub_dt.0;
+            let new_ang_vel = (rot.mul(prev_rot.inverse())).as_radians() / sub_dt.0;
+            // avoid triggering bevy's change detection unnecessarily
+            if new_ang_vel != ang_vel.0 {
+                ang_vel.0 = new_ang_vel;
+            }
         }
     }
 }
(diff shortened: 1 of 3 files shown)


RJ:
adds some verbosity, happy to make any style / naming changes or whatever you suggest.

src/plugins/solver.rs
```diff
@@ -373,12 +384,34 @@ fn solve_vel(
...
+                    body1.linear_velocity.0 += lin_vel_1;
+                }
+                #[cfg(feature = "2d")]
+                if ang_vel_1 != 0.0 {
+                    body1.angular_velocity.0 += ang_vel_1;
+                }
+                #[cfg(feature = "3d")]
+                if ang_vel_1 != Vector::ZERO {
+                    body1.angular_velocity.0 += ang_vel_1;
+                }
```
Jondolf, line 399: - Maybe use `delta_lin_vel`/`delta_ang_vel` here since it's what is used earlier
- I'm not sure if this is more or less readable, but the 2D/3D could technically be combined with `AngularVelocity::ZERO.0`
```suggestion
                let delta_lin_vel = p * inv_mass1;
                let delta_ang_vel = compute_delta_ang_vel(inv_inertia1, r1, p);

                if delta_lin_vel != Vector::ZERO {
                    body1.linear_velocity.0 += delta_lin_vel;
                }
                if delta_ang_vel != AngularVelocity::ZERO.0 {
                    body1.angular_velocity.0 += delta_ang_vel;
                }
```

src/plugins/solver.rs
```diff
@@ -373,12 +384,34 @@ fn solve_vel(
...
+                    body2.linear_velocity.0 += lin_vel_2;
+                }
+                #[cfg(feature = "2d")]
+                if ang_vel_2 != 0.0 {
+                    body2.angular_velocity.0 += ang_vel_2;
+                }
+                #[cfg(feature = "3d")]
+                if ang_vel_2 != Vector::ZERO {
+                    body2.angular_velocity.0 += ang_vel_2;
+                }
```
Jondolf, line 414: - Same as above
- The delta should be subtracted, not added (Newton's 3rd law, equal and opposite force)
```suggestion
                let delta_lin_vel = p * inv_mass2;
                let delta_ang_vel = compute_delta_ang_vel(inv_inertia2, r2, p);

                if delta_lin_vel != Vector::ZERO {
                    body2.linear_velocity.0 -= delta_lin_vel;
                }
                if delta_ang_vel != AngularVelocity::ZERO.0 {
                    body2.angular_velocity.0 -= delta_ang_vel;
                }
```

Jondolf (changes requested):
I left some comments on style + an error in the velocity solver, otherwise this looks good!

src/plugins/solver.rs
RJ: ah good yeah i like`AngularVelocity::ZERO.0` better than the cfg flag

Jondolf (approved):

PR #174, opened 2023-10-12
Ensure newly created bodies get simulated the frame they are added

RJ:
### First commit

Added an apply_deferred to ensure all the physics components get added that frame - otherwise the body won't be simulated until commands are flushed later.

### Second commit

HOWEVER, the added test will still fail because it appears that delta_seconds is 0 for the first update here:
https://github.com/RJ/bevy_xpbd/blob/c61a90ec03ba51b74354053eca9ac8c7f09d1f59/src/plugins/setup.rs#L260

meaning the ` physics_loop.accumulator` doesn't increase on the first update, because it was referencing `delta_seconds`. I *think* it makes more sense to use `dt` there, since it's already been scaled and clamped as needed.

This fixes the test, but only because it's not using a Variable timestep, which would still fail because there is no delta_seconds on the first update. I don't think it's the fault of the manual time updating strategy; i think delta_seconds just isn't calcualted on the first ever update: https://github.com/bevyengine/bevy/blob/bb13d065d36c9b495b755aef2040b43f29a91b79/crates/bevy_time/src/time.rs#L155

It's a bit of a weird edge case, and as long as the tests don't want to use variable timesteps and have things simulated on the first ever update, it hopefully won't be an issue in future.

### DIFF
--- a/crates/bevy_xpbd_2d/snapshots/bevy_xpbd_2d__tests__body_with_velocity_moves.snap
+++ b/crates/bevy_xpbd_2d/snapshots/bevy_xpbd_2d__tests__body_with_velocity_moves.snap
@@ -4,7 +4,7 @@ expression: transform
 ---
 Transform {
     translation: Vec3(
-        8.316667,
+        8.35,
         0.0,
         0.0,
     ),
--- a/crates/bevy_xpbd_3d/snapshots/bevy_xpbd_3d__tests__body_with_velocity_moves.snap
+++ b/crates/bevy_xpbd_3d/snapshots/bevy_xpbd_3d__tests__body_with_velocity_moves.snap
@@ -4,7 +4,7 @@ expression: transform
 ---
 Transform {
     translation: Vec3(
-        8.316667,
+        8.35,
         0.0,
         0.0,
     ),
--- a/src/plugins/prepare.rs
+++ b/src/plugins/prepare.rs
@@ -58,6 +58,8 @@ impl Plugin for PreparePlugin {
                 init_colliders,
                 update_mass_properties,
                 clamp_restitution,
+                // all the components we added above must exist before we can simulate the bodies
+                apply_deferred,
             )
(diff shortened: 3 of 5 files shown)


Jondolf:
I think using `dt` instead of `delta_seconds` is very different though and breaks the time accumulator. `delta_seconds` is the "real time" while `dt` is the time indicated by the `PhysicsTimestep` configuration, i.e. `1.0 / 60.0` by default with a fixed timestep. If you add `dt` to the accumulator, you're not actually accumulating the real elapsed time.

RJ:
Hm I see, ok.

Presumably we still expect the physics accumulator to increase on the first call to app.update(), so can we do something like this?

```rust
    // Update `DeltaTime` according to the `PhysicsTimestep` configuration
    let (raw_dt, accumulate) = match time_step {
        PhysicsTimestep::Fixed(fixed_delta_seconds) => (fixed_delta_seconds, true),
        PhysicsTimestep::FixedOnce(fixed_delta_seconds) => (fixed_delta_seconds, false),
        PhysicsTimestep::Variable { max_dt } => (delta_seconds.min(max_dt), true),
    };

    // case where first ever call to app.update()
    if world.resource::<Time>().last_update().is_none() {
        delta_seconds = raw_dt;
    }
```

Doesn't help the Variable case, but i'm not sure what to do about that. There's time.first_update() when the app started up, but i don't think it makes sense to just take the difference for the variable case - otherwise the delta includes bevy boot up time.

Jondolf:
> Presumably we still expect the physics accumulator to increase on the first call to app.update(), so can we do something like this?

Yep, this is what I had in mind as well. As for the `Variable` case, I don't think we need to do anything about it; by definition, it advances the simulation by the elapsed real time, and on the first frame there is no elapsed time.

Jondolf:
Updated Insta snapshots to make CI pass (hopefully)

PR #192, opened 2023-10-22
Map entities

johanhelsing:
# Objective

For bevy_ggrs and other snapshot/save/load/scenes functionality to work the `MapEntities` trait is needed in order to fix entity ids after loading.

## Solution

Implement `MapEntities` for the components that have `Entity`s as fields.

---

## Changelog

- Components now implement `MapEntities`

## Migration Guide

- The `XpbdConstraint` trait now requires `MapEntities` if you implemented custom `XpbdConstraint`s you should now implement the `MapEntities` trait on your types.

### DIFF
--- a/src/constraints/joints/mod.rs
+++ b/src/constraints/joints/mod.rs
@@ -96,10 +96,10 @@ pub use revolute::*;
 pub use spherical::*;
 
 use crate::prelude::*;
-use bevy::prelude::*;
+use bevy::{ecs::entity::MapEntities, prelude::*};
 
 /// A trait for [joints].
-pub trait Joint: Component + PositionConstraint + AngularConstraint {
+pub trait Joint: Component + PositionConstraint + AngularConstraint + MapEntities {
     /// Creates a new joint between two entities.
     fn new(entity1: Entity, entity2: Entity) -> Self;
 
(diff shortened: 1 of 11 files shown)


src/constraints/joints/mod.rs
```diff
@@ -96,10 +96,10 @@ pub use revolute::*;
 pub use spherical::*;
 
 use crate::prelude::*;
-use bevy::prelude::*;
+use bevy::{ecs::entity::MapEntities, prelude::*};
 
 /// A trait for [joints].
-pub trait Joint: Component + PositionConstraint + AngularConstraint {
```
Jondolf, line 102: If we want to add `MapEntities` to the trait bounds, would it make sense to have it on `XpbdConstraint` directly? This way it'd be required for all constraints and not just joints, and all constraints currently have entities anyway

Jondolf:
Looks like it needs to be implemented for `CustomDistanceConstraint` in the `custom_constraint` example now :P

Jondolf (approved):
Thanks, looks good now :)

johanhelsing:
Btw, I've been working on getting bevy_gaff to work. Turned out desync detection in ggrs was broken as well as these various small non-determinism sources.

Still trying to get a single particle without collisions and just force as input to work.

...but hopefully not too far away now.

PR #204, opened 2023-10-28
[bevy-0.12] Migrate to Interned ScheduleLabels

rodolphito:
For Bevy 0.12

the only sketchy thing here is `setup.rs:128`
```rs
let fixed_update = schedule == FixedUpdate.intern();
```


Jondolf:
Thanks! CI is failing, but it's probably because I haven't done the rest of the migration properly yet, and some things seem to have changed on Bevy main. I'll do it today though

Jondolf:
I'll merge this since I need to do this anyway, and fix the remaining issues (+ verify the fixed update check) in the base migration PR

Jondolf:
Btw I tested and the fixed update check does seem to work, it's consistently returning `true` when running physics in `FixedUpdate` and `false` otherwise

PR #216, opened 2023-11-01
docs: Use more backticks.

waywardmonkeys:
# Objective

- Don't have warnings from `clippy::doc_markdown`.
- Have nicely formatted API docs.

## Solution

- Use backticks more!


Jondolf:
Thanks, didn't know about `clippy::doc_markdown`, looks very useful! Maybe you could also add it as a crate-level warning in `lib.rs` like `#![warn(clippy::doc_markdown)]`? I'm not getting the warnings without it, and it's nice to have CI catch it as well

waywardmonkeys:
Pushed a commit enabling it.

PR #219, opened 2023-11-02
Move x comparison in broad phase up to early-out as soon as possible

datael:
# Objective

- At around 13,000 colliders, none of which could collide with each other, I was seeing the broad phase alone taking ~20ms
- Wanting to bring it down to something more manageable, I experimented with various ways to potentially speed up the broad phase

## Solution

- Believe it or not, the biggest gain I found was to simply break out of the loop sooner
    - Move the `aabb2.mins.x > aabb1.maxs.x` check that can break out of the loop to the very top, before the inactive/layer checks
    - This resulted in the 13,000 colliders going from taking ~20ms to just ~2ms
    - I've understood this to be caused by the layer check causing the `continue` to be triggered, preventing the loop from ever hitting the early `break` branch


datael:
Before:

Screenshot 2023-11-02 at 22 30 06

After:

Screenshot 2023-11-02 at 22 30 25

Jondolf (approved):
Thanks, good catch!

Issue #224, opened 2023-11-05
Collision events are sent when not actually in contact


Jondolf:
Collision events are currently sent for contacts where the distance is below `NarrowPhaseConfig.prediction_distance`, but not necessarily touching. This can lead to unexpected event triggers as reported by `@COCPORN` in a situation where a ball sometimes bounces off of an object that despawns on contact and sometimes not.

Collision events should only be sent for contacts that are actually touching.

For now, you can get around the issue with this:

```rust
.insert_resource(NarrowPhaseConfig {
    prediction_distance: 0.0,
})
```

but it can cause contact stability issues in cases where moving bodies bump into each other.

Technostalgic:
I also just want to say that it is actually quite nice to be able to query for "almost" collisions, like how it works now. Of course this should be fixed but there is also definitely a use case for the way that it works currently, as well. So maybe an unwarranted suggestion but maybe when this does get fixed, also add in a way to query for potential collisions within the prediction distance too?

Some use cases:

* Ground detection: It's nice to have a bit of leeway when performing ground detection, I personally just use contacts and coyote time, but I know raycasting is also a common method to look for ground detection, and this would be a similar way to do it as shape/ray casting, but better since you don't have to do an additional query, and it covers the exact shape of the colldier.  

* Wall Detection: Sometimes when directly next to a wall, if there isn't some force pressing you directly up against the wall, there will be no collision between the target body and the wall since the solver has moved the body to no longer be intersecting with the wall.

Jondolf:
This should be fixed by #385. For those use cases I would still recommend using spatial queries like ray casts or shape casts, or even sensors that act as the "enlarged" collider.

PR #232, opened 2023-11-10
docs: Point users to `NarrowPhaseConfig::prediction_distance`

johanhelsing:
# Objective

I tripped up on collisions, because I considered the units as meters, but the default value for prediction distance is 1 (presumably pixels)

## Solution

Most of the documentation links to the `ContactReportingPlugin` when talking about contacts.

So link from there to `NarrowPhaseConfig::prediction_distance`, so they can follow the trail.

### DIFF
--- a/src/plugins/collision/contact_reporting.rs
+++ b/src/plugins/collision/contact_reporting.rs
@@ -17,6 +17,9 @@ use crate::prelude::*;
 ///
 /// You can listen to them with normal event readers:
 ///
+/// Colliders are considered as colliding if they are closer than
+/// [`NarrowPhaseConfig::prediction_distance`] away from each other.
+///
 /// ```no_run
 /// use bevy::prelude::*;
 #[cfg_attr(feature = "2d", doc = "use bevy_xpbd_2d::prelude::*;")]


Jondolf:
The recommended units are currently pixels in 2D and meters in 3D, which applies to the prediction distance as well. Once we add physics scale like `pixels_per_meter`, 2D can perhaps be handled more conveniently.

I'm a bit unsure if we want to have the part about bodies being considered colliding with the penetration distance in the docs though, because I consider the behavior to be somewhat incorrect. The prediction distance is used just to avoid missing collisions; the narrow phase collects all contacts before the solver starts, and the constraints can later move bodies around into penetrating states, which would only be detected during the next frame without the extra margin. It only causes a collision response if the bodies are actually penetrating though, so people would expect it to skip sending collision events for non-penetrating collisions.

I made an issue about this (#224) and will try to fix it for 0.4

johanhelsing:
Alright, just wanted make sure it's possible to figure out how to use the engine correctly...

Are you sure it's not simpler to just use meters everywhere?

- get rids of the weirdness surrounding the value of default gravity
- the actual values of mass, velocities, forces etc. will be in a more natural range that's easier for humans to reason about.
- Not all games are pixel-based either. 3D games with 2d physics are not uncommon either.
- Matches the conventions in other physics engines (I think, at least Unity/box2d)
- 2D and 3D have less differences/special cases

Jondolf:
Closing this, as the property doesn't even exist in this form anymore after the solver rework (#385), and the collision event issue (#224) has also been fixed.

I do agree that MKS (meters, kilograms, seconds) should generally be preferred, and in 2D you could do that by simply expressing object sizes in meters and scaling the camera projection to the desired scale. The new solver in fact uses meters for length-based tolerances and such.

We can't force everyone to use meters though, apart from recommending it in docs, and in some cases pixels might just be easier to use. For this, the new solver has a [`PhysicsLengthUnit`](https://docs.rs/avian2d/latest/avian2d/dynamics/solver/struct.PhysicsLengthUnit.html) resource to allow the engine to tune its tolerances to match the expected scale. Rapier, Box2D V3, and PhysX all have a similar configuration option.

PR #237, opened 2023-11-10
Normalize rotation after applying constraint delta

### LINKED ISSUE #235: rotations not normalized in solver, causing colliders to expand and eventually explode
image

Adding  `*body1.rotation.0 = *body1.rotation.0.normalize();` [after rotational updates in `position_constraints.rs`](https://github.com/Jondolf/bevy_xpbd/blob/776a785c8b2accacc40dcb4a84c1a72ffe6890bb/src/constraints/position_constraint.rs#L33) fixes  this.  (Doesn't seem to be necessarily in angular_constraints.rs, but I think that's because it happens that the constraints are applied with positional_constraints running after in all existing joints which use both — so that should also be done just in case.)

I think rather than sticking the normalization after every rotation, it could be done by refactoring [`get_delta_rot()`](https://github.com/Jondolf/bevy_xpbd/blob/776a785c8b2accacc40dcb4a84c1a72ffe6890bb/src/constraints/position_constraint.rs#L94) —  this would mean less code duplication, but probably also it getting called a few more times than necessary. Maybe it's actually better to put it _up_ into the solver step somewhere, so it only gets called once.

If joints end up getting unified into one 6dof universal joint, that might change this entirely. So, I'm going to submit a PR for quick-and-easy fix now, but am also filing this so we can later look at putting the normalization at the most optimal level.

mattdm:
# Objective

Fixes #235 

## Solution

Put `bodyN.rotation.0.normalize();` after every use of `Self::get_delta_rot` in position_constraints and angular_constraints.

THis isn

---

## Changelog

- rotating colliders involved in collisions much less likely to make them explode

### DIFF
--- a/src/constraints/angular_constraint.rs
+++ b/src/constraints/angular_constraint.rs
@@ -67,9 +67,11 @@ pub trait AngularConstraint: XpbdConstraint<2> {
         // Apply rotational updates
         if body1.rb.is_dynamic() {
             *body1.rotation += Self::get_delta_rot(rot1, inv_inertia1, p);
+            *body1.rotation.0 = *body1.rotation.0.normalize();
         }
         if body2.rb.is_dynamic() {
             *body2.rotation -= Self::get_delta_rot(rot2, inv_inertia2, p);
+            *body2.rotation.0 = *body2.rotation.0.normalize();
         }
 
         p
--- a/src/constraints/position_constraint.rs
+++ b/src/constraints/position_constraint.rs
@@ -34,10 +34,21 @@ pub trait PositionConstraint: XpbdConstraint<2> {
         if body1.rb.is_dynamic() && body1.dominance() <= body2.dominance() {
             body1.accumulated_translation.0 += p * inv_mass1;
             *body1.rotation += Self::get_delta_rot(rot1, inv_inertia1, r1, p);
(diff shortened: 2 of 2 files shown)


src/constraints/angular_constraint.rs
```diff
@@ -67,9 +67,11 @@ pub trait AngularConstraint: XpbdConstraint<2> {
         // Apply rotational updates
         if body1.rb.is_dynamic() {
             *body1.rotation += Self::get_delta_rot(rot1, inv_inertia1, p);
+            *body1.rotation.0 = *body1.rotation.0.normalize();
```
Jondolf, line 70: `rotation.0` is the quat itself, so it can't be dereferenced, which is making CI fail. Same in other places
```suggestion
            body1.rotation.0 = body1.rotation.normalize();
```

mattdm: Hmmm. I did that _for some reason_ that I don't remember now. I'll check and fix the PR. Thanks!

mattdm:
Sorry for vanishing! I plan to come back to working on this in January.

Jondolf:
Hi, I've adopted this PR in #345 since issues related to this have come up again. I'll add you as a co-author though :)

PR #241, opened 2023-11-12
Filter collisions between children of the same rigidbody in broad phase

mbrea-c:
# Objective

Currently, collisions between multiple child colliders of the same rigidbody are ignored by the solver. However:

- Collision events are generated for such cases.
- All such collisions are processed in the narrow phase, which has a large performance impact in projects that rely heavily on child colliders.

Preferably all such collisions should be filtered out earlier, in the broad phase, to avoid both of these issues.

## Solution

- Keep the `ColliderParent` of a collider in `AabbIntervals` in the broad phase.
- Ignore a collision if the `ColliderParent` of two entities is equal.

## Changelog

- Collisions between children of the same rigidbody are no longer reported in contact events and are no longer processed in the narrow phase.

### DIFF
--- a/src/plugins/collision/broad_phase.rs
+++ b/src/plugins/collision/broad_phase.rs
@@ -158,31 +158,43 @@ type IsBodyInactive = bool;
 
 /// Entities with [`ColliderAabb`]s sorted along an axis by their extents.
 #[derive(Resource, Default)]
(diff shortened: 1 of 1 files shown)


Jondolf (approved):
LGTM, thanks!

Issue #244, opened 2023-11-13
Unstable when multiple colliders spawned as rigid body children


dafteran4:
This is a simple example using multiple colliders spawned as children for one dynamic rigid body. It produces exploding / unstable behavior. Version is 0.3.2. It is also unstable if only one children is spawned. It is, however, fine if collider is part of the same entity as dynamic body.

```
use bevy::prelude::*;
use bevy_xpbd_3d::prelude::*;

fn main() {
    App::new()
        .add_plugins((
                DefaultPlugins,
                PhysicsPlugins::default(),
                PhysicsDebugPlugin::default(),
        ))
        .add_systems(Startup, test)
        .run();
}

fn test(mut commands: Commands) {

    let n = 10;
    let size = 20.;

    let mut heights = Vec::new();
    heights.resize_with(n, || {
        let mut v = Vec::new();
        v.resize(n, 0.);
        v
    });

    commands.spawn((
            Collider::heightfield(heights, Vec3::splat(size)),
            RigidBody::Static,
    ));

    commands
        .spawn((RigidBody::Dynamic, SleepingDisabled))
        .with_children(|c| {
            c.spawn((Collider::capsule(1.0, 0.1), Transform::from_xyz(0.0, 1.0, 0.0), ColliderDensity(100.)));
            c.spawn((Collider::capsule(1.0, 0.1), Transform::from_xyz(0.05, 2.0, 0.0), ColliderDensity(200.)));
        }); 

    commands.insert_resource(Gravity(Vec3::NEG_Y * 9.8));

    commands.spawn(Camera3dBundle {
        transform: Transform::from_xyz(0., 2., 7.).looking_at(Vec3::ZERO, Vec3::Y),
        ..default()
    });
    
}


```

dafteran4:
Also tested this on main branch 6444910 , it behaves the same.

mbrea-c:
The rigidbody has an offset center of mass so the problem might be related to #246. I'm working on a PR but it's running into some issues atm.

Jondolf:
Should be fixed by #250.

dafteran4:
Thanks. It works now.

PR #250, opened 2023-11-15
Fix rotations when center of mass is offset

### LINKED ISSUE #246: Rotational motion ignores center of mass
### Problem
When given an angular velocity, rigidbodies that have a non-zero local center of mass seem to rotate around their `Position`, ignoring `CenterOfMass`:

https://github.com/Jondolf/bevy_xpbd/assets/31567043/9050e518-7ff3-4e2f-b50f-a9fd3d96d7dc

Code for reproducing:
```rust
    // floor
    commands
        .spawn(PbrBundle {
            mesh: meshes.add(shape::Box::new(40., 0.3, 40.).into()),
            material: materials.add(Color::WHITE.into()),
            ..default()
        })
        .insert(RigidBody::Static)
        .insert(Collider::cuboid(40., 0.3, 40.));

    // cube
    commands
        .spawn(SpatialBundle {
            transform: Transform {
                translation: Vec3::new(0.0, 2.0, 0.0),
                ..default()
            },
            ..default()
        })
        .insert(RigidBody::Dynamic)
        .insert(AngularVelocity(Vec3::new(0., 7., 0.)))
        .insert(LinearVelocity(Vec3::new(0., 7., 0.)))
        .with_children(|b| {
            b.spawn(SpatialBundle {
                transform: Transform {
                    translation: Vec3::new(1.5, 0., 0.),
                    ..default()
                },
                ..default()
            })
            .insert(Collider::cuboid(1., 1., 1.));
        });
```

### Possible solution

First, is this unintended or am I missing something? If it's unintended, we might want to account for the center of mass during integration. For reference, this is what rapier does: [link](https://github.com/dimforge/rapier/blob/master/src/dynamics/rigid_body_components.rs#L566)

Updating `integrate_rot` in `integration.rs` and `update_lin_vel` in `solver.rs` to account for offset centers of mass makes the rotation work as expected, but seems to introduce a number of other issues (exploding velocity on contact, object suddenly speeding up when adjusting timescale). Any suggestions on how best to fix this?

thanks :)

mbrea-c:
# Objective

Fixes #246 and #244.

## Solution

If we consider `AccumulatedTranslation` to represent the translation of the center of mass (instead of the position),
it's only necessary to account for the center of mass offset when applying the translation. Since we already have access to the original position and translation, computing the change in `Position` based on the change in center of mass position and rotation is straightforward.

Usages of `AccumulatedTranslation` that assumed it refers to a translation applied directly to `Position` had to be updated. Fortunately, `AccumulatedTranslation` is often accessed through `RigidBodyQuery::current_position`, so not many changes were needed.

## Performance

There's no measurable performance impact when testing in my game with hundreds of child colliders and profiling with Tracy (external trace includes this PR, this trace is main branch):

performance

## Alternative solutions considered

An alternative approach is to apply the additional translation due a rotation every time we apply a rotation (during integration,  constraint solving) and take into account only the change in position of the center of mass in `update_lin_vel`. This was what I tried first, but I was unable to get it to work due to numerical inaccuracies that caused unexpected accelerations, or some other insidious bug I couldn't find :). Either way, this alternative would have required more changes, and for what it's worth I personally found it less clean as there was no longer a clean separation between updates to position and rotation (every rotation had to add a small change to the position).

## Changelog

- `AccumulatedTranslation` now refers to the translation of the center of mass, which may not equal the translation applied to `Position` when the center of mass is offset.
- Rotational motion with offset centers of mass now works as expected.
  - As a result, child colliders finally *feel* stable in my testing.

### DIFF
--- a/src/components/world_queries.rs
+++ b/src/components/world_queries.rs
@@ -1,6 +1,6 @@
 #![allow(missing_docs)]
 
-use crate::prelude::*;
+use crate::{prelude::*, utils::get_pos_translation};
 use bevy::ecs::query::WorldQuery;
 use std::ops::{AddAssign, SubAssign};
 
@@ -69,7 +69,13 @@ impl<'w> RigidBodyQueryItem<'w> {
     /// Returns the current position of the body. This is a sum of the [`Position`] and
     /// [`AccumulatedTranslation`] components.
     pub fn current_position(&self) -> Vector {
-        self.position.0 + self.accumulated_translation.0
+        self.position.0
+            + get_pos_translation(
+                &self.accumulated_translation,
+                &self.previous_rotation,
+                &self.rotation,
+                &self.center_of_mass,
+            )
     }
 
     /// Returns the [dominance](Dominance) of the body.
--- a/src/plugins/solver.rs
+++ b/src/plugins/solver.rs
@@ -584,6 +584,7 @@ pub fn joint_damping<T: Joint>(
     }
 }
 
+#[allow(clippy::type_complexity)]
 fn apply_translation(
     mut bodies: Query<
         (
--- a/src/plugins/solver.rs
+++ b/src/plugins/solver.rs
@@ -602,10 +602,9 @@ fn apply_translation(
             continue;
         }
 
-        let prev_global_com = pos.0 + prev_rot.0.rotate(center_of_mass.0);
-        let next_global_com = prev_global_com + translation.0;
-
-        pos.0 = next_global_com - rot.rotate(center_of_mass.0);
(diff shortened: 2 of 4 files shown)


Jondolf (approved):
Huge thanks for looking into this! The behavior looks a lot more realistic and stable with your fixes and resolves a lot of issues with child colliders. I'm a bit surprised that it's as easy as applying `AccumulatedTranslation` a bit differently. Thanks again!

mbrea-c:
No problem, thank you for your work on this project!

PR #255, opened 2023-11-18
Use any_orthogonal_vector to get orthogonal vector

ollef:
Using Vector::Y here doesn't work if axis1 is parallel with Y (or close to it due to precision issues).


Jondolf (approved):
Thanks!

Issue #256, opened 2023-11-19
[feature] Support for child entities enacting force on parent `RigidBody`


ActuallyHappening:
Hello and this crate is awesome!

I am trying to build a space-themed game using `bevy 0.12` which revolves heavily around player movement (no gravity). The ship entity is a `RigidBody::Dynamic` but has no `Collider` (does have `MassPropertiesBundle`), gaining its shape *and movement forces* from child entities like blocks, thrusters or weapons.
To move the ship, initially I placed an `ExternalForce` on a child entity (thruster), but nothing moved. Then I made the thruster (the child entity) a `RigidBody::Dynamic / Static / Kinematic` but the child entity itself moved (and collided into other blocks making up the ship) which is not the behaviour I was looking for.

Is there an existing API for other entities enacting forces on parent entities?

From the assumption there is not, I propose a new component: `ExternalForceParent` (and variants for impulse and torque). Then, theoretically a single system `update_parent_forces` can iterate through all components with `ColliderParent` and `ExternalForceParent`, assert the parent's `ExternalForce`.persistence is false, and apply the force with `.apply_force_at_point`.
I see problems with this approach, however. Does `ExternalForce`.apply_force_at_point take into account the centre of mass? Is this needlessly complex, adding three similar extra types? Why can't we re-use `ExternalForce` somehow in child entities?

I would love to here your thoughts, especially if I am missing some existing API

Jondolf:
`ExternalForceParent` feels too niche for me. There are other ways to achieve the desired behavior, and I don't like introducing a ton of new yet similar components for something so specific.

`apply_force_at_point` is given the local center of mass. So yes, it should take it into account, assuming the given center of mass is correct.

In your case, I would either manually implement the logic that accumulates the child forces to the parent's `ExternalForce`, taking into account their offsets, or use joints. You could use a `DistanceJoint` or some other joint type to align the thrusters etc. to specific offsets (with local anchor points) relative to the actual ship. This way, applying forces on the thrusters (and other connected bodies) should automatically apply forces and torques to the ship correctly.

ActuallyHappening:
For my specific use case, I manually implemented a solution here: https://github.com/ActuallyHappening/bevy_xpbd3d_parenting
It does rely on `clear_forces_and_impulses` being public, which is the reason for #257 
[...]

PR #257, opened 2023-11-20
fix: make `clear_forces_and_impulses` public

ActuallyHappening:
# Objective

I am wanting to implement custom logic that runs just after `ExternalForce` is cleared. This helps solve issue #256.

## Solution

Make the function `clear_forces_and_impulses` public, so that other external systems can order against it. In my case, I want to order `.after(clear_forces_and_impulses)`.

---

### Changed
The function `clear_forces_and_impulses was make public`


src/components/mass_properties.rs
```diff
@@ -193,6 +193,11 @@ pub struct CenterOfMass(pub Vector);
 impl CenterOfMass {
     /// A center of mass set at the local origin.
     pub const ZERO: Self = Self(Vector::ZERO);
+
+		/// Get the inner [`Vector`] in local space of the [`CenterOfMass`]
+		pub fn inner(&self) -> Vector {
+			self.0
+		}
```
Jondolf, line 200: This feels very unnecessary, you can just do `center_of_mass.0` like with all other components

src/plugins/integrator.rs
```diff
@@ -295,7 +295,11 @@ type ForceComponentsChanged = Or<(
     Changed<ExternalAngularImpulse>,
 )>;
 
-fn clear_forces_and_impulses(mut forces: Query<ForceComponents, ForceComponentsChanged>) {
+
```
Jondolf, line 298: CI linting doesn't like double line breaks
```suggestion
```

src/plugins/integrator.rs
```diff
@@ -310,4 +314,4 @@ fn clear_forces_and_impulses(mut forces: Query<ForceComponents, ForceComponentsC
             angular_ímpulse.clear();
         }
     }
-}
+}
```
Jondolf, line 317: Files need to have a trailing newline
```suggestion
}

```

src/plugins/integrator.rs
```diff
@@ -295,7 +295,11 @@ type ForceComponentsChanged = Or<(
     Changed<ExternalAngularImpulse>,
 )>;
 
-fn clear_forces_and_impulses(mut forces: Query<ForceComponents, ForceComponentsChanged>) {
+
+/// Responsible for clearing forces and impulses on bodies.
+/// 
+/// Public so external crates can order systems after this, runs in [`PhysicsSchedule`], after [`PhysicsStepSet::SpatialQuery`]
```
Jondolf, line 301: Justifying the publicity here feels unnecessary since it applies to all public systems
```suggestion
/// Responsible for clearing forces and impulses on bodies.
/// 
/// Runs in [`PhysicsSchedule`], after [`PhysicsStepSet::SpatialQuery`].
```

src/components/mass_properties.rs
ActuallyHappening: Yes I agree it is unnecessary
I do like the ability to put documentation on it though, but will remove at your request

ActuallyHappening:
OK I believe all changes are in order

src/plugins/integrator.rs
```diff
@@ -295,7 +295,10 @@ type ForceComponentsChanged = Or<(
     Changed<ExternalAngularImpulse>,
 )>;
 
-fn clear_forces_and_impulses(mut forces: Query<ForceComponents, ForceComponentsChanged>) {
+/// Responsible for clearing forces and impulses on bodies.
+///
```
Jondolf, line 299: Looks like there's some whitespace here
```suggestion
///
```

ActuallyHappening: Damn my bad

PR #279, opened 2023-12-15
Support multiple integrators and integrating outside ECS

Jondolf:
# Objective

Currently, only semi-implicit Euler integration is supported, and it can only be done through the integration systems. It would be useful to have alternative integration schemes, both to support more use cases and to compare the pros and cons of each method.

The integration schemes should also be simple functions instead of systems so that they can be used without access to the ECS. This makes testing easier, allows e.g. predicting positions in a simple way, and makes it trivial to support different integration schemes for different bodies. For example, cloth particles might want to use Verlet integration while rigid bodies should use semi-implicit Euler integration.

## Solution

- Clean up the current integration logic and move it into its own functions in a `semi_implicit_euler` module
- Add `linear_acceleration` and `angular_acceleration` helpers
- Add the velocity Verlet integration scheme
- Add an `IntegrationScheme` enum resource for changing the default solver
- Add tests and docs

## Future work

- [ ] Add an example that renders curves showing the analytical result and the result produced by our integrators
- [ ] Benchmarks
- [ ] More integration schemes:
  - Explicit Euler integration (usually just worse than semi-implicit, but useful for comparison)
  - Implicit Euler integration (more expensive than explicit and semi-implicit but very accurate)
  - Runge-Kutta 4, aka. RK4 (relatively expensive but very accurate, useful when high precision is required)
  - Other, not very useful
    - Basic Verlet integration (equivalent to velocity Verlet, but uses current and previous position, doesn't support specifying velocity)
    - Leapfrog (equivalent to velocity Verlet, but shifted by half a timestep)


Jondolf:
Closing this for now; while supporting multiple integration schemes could be cool, it's not as useful in practice, and some integration methods also just don't work as well with the solver. Symplectic Euler is just generally the best option for games.

I will rework the integration code a bit for the new upcoming impulse-based solver though, as it requires position and velocity integration to be done in separate systems.

PR #283, opened 2023-12-27
Ignore static-static collisions in broad phase

Jondolf:
# Objective

Currently, the broad phase determines whether a body is inactive just by checking if the position or rotation has changed. If a static body is moved, it will unnecessarily check collisions against other static bodies. This can cause performance issues with e.g. floating point worlds where all entities are frequently moved.

## Solution

Also check whether the body is static in the `update_aabb_intervals` system.

### DIFF
--- a/src/plugins/collision/broad_phase.rs
+++ b/src/plugins/collision/broad_phase.rs
@@ -184,6 +184,7 @@ impl MapEntities for AabbIntervals {
 #[allow(clippy::type_complexity)]
 fn update_aabb_intervals(
     aabbs: Query<(
+        &RigidBody,
         &ColliderAabb,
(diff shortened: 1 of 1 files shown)


Plash-har:
`update_aabb_intervals` should treat every collider ? Because if a collider doesn't have a `RigidBody` component but the parent does (and the parent doesn't possess any `Collider`) then this collider isn't treated by this system ?

Jondolf:
Ah this should check the `ColliderParent` and use that `RigidBody` instead; I'll fix that

Jondolf:
Should be fixed by #285, thanks for reporting the issue. A dumb oversight on my part :P

PR #292, opened 2024-01-01
Make PreparePlugin configurable

Rigidity:
# Objective

It's important to be able to disable `transform->position` and `position->transform` syncing in the `PreparePlugin` so that positions don't get messed up if using an alternative coordinate system, such as a floating origin. This PR makes `PreparePlugin` more configurable and exposes a more detailed `PrepareSet` to make external customization easier to manage.

## Solution

Adds `position_to_transform` and `transform_to_position` fields to a new `PrepareConfig`, and exposes `PrepareSet` to the public so that you can add your own systems there without leaking too many implementation details.

---

### Added
- Added `PrepareConfig`, based off of `SyncConfig`.

### Changed
- Split up `PrepareSet` into individual groups.
- Refactored the system scheduling in `PreparePlugin`.
- Made `init_transforms` respect `PrepareConfig`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -177,6 +177,7 @@
 //!     - [`PhysicsSchedule`] and [`PhysicsStepSet`]
 //!     - [`SubstepSchedule`] and [`SubstepSet`]
 //!     - [`PostProcessCollisions`] schedule
+//!     - [`PrepareSet`]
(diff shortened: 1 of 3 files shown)


Jondolf (approved):
Reviewed this finally; The new structure looks good, and I like the documentation a lot. This adds some useful flexibility and configurability. Thanks!

PR #297, opened 2024-01-12
Implement cast_ray_predicate to allow filtering the colliders with a function

Affinator:
As I was writing a xpbd backend for bevy_mod_picking, I found that a ray cast with a predicate was missing to implement the functionality efficiently. 

I was inspired by the predicate from bevy_rapier: https://github.com/dimforge/bevy_rapier/blob/c6bcce4695d596a7a9c8e91748d4dbb3d31f6d13/src/pipeline/query_filter.rs#L12

# Objective

- Implement a ray cast function with the support for a predicate function to filter the colliders

## Solution

- Added support for a predicate starting at SpatialQuery and down the pipeline
- I added a small (and fun to play with) example to illustrate one possible use case
- I chose not to change existing functionality. Therefore there is some code duplication in QueryPipelineAsCompositeShapeWithPredicate. This could be unified by introducing an Option in QueryPipelineAsCompositeShape, but that would create breaking changes at many places.

### DIFF
--- a/crates/bevy_xpbd_3d/examples/cast_ray_predicate.rs
+++ b/crates/bevy_xpbd_3d/examples/cast_ray_predicate.rs
@@ -1,6 +1,6 @@
 #![allow(clippy::unnecessary_cast)]
 
-use bevy::{prelude::*, pbr::NotShadowReceiver};
+use bevy::{pbr::NotShadowReceiver, prelude::*};
 use bevy_xpbd_3d::{math::*, prelude::*};
 use examples_common_3d::XpbdExamplePlugin;
 
@@ -21,7 +21,6 @@ struct MovementAcceleration(Scalar);
 /// If to be ignored by raycast
 #[derive(Component)]
 struct OutOfGlass(bool);
-
 
 const CUBE_COLOR: Color = Color::rgba(0.2, 0.7, 0.9, 1.0);
 const CUBE_COLOR_GLASS: Color = Color::rgba(0.2, 0.7, 0.9, 0.5);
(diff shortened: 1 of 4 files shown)


Affinator:
My branch should now pass all workflows (and I updated it to the current state of main).

Affinator:
Should i squash the commits or is this fine the way it is?
(Or are there other concerns?)

Jondolf:
Sorry for responding so late.

I was originally considering a predicate in `SpatialQueryFilter` like what Rapier does, but I think that would have annoying lifetime issues when used with the `RayCaster`/`ShapeCaster` components, so your approach is definitely fine too.

Note that predicates should already work when using `ray_hits_callback`, but having one for `cast_ray` is definitely useful. I'd like to have `cast_shape_predicate` and `project_point_predicate` too if you have time, but leaving them to a follow-up is fine as well.

The example is really fun, but it's a bit confusing in that it only colors one box even though the laser/indicator goes through multiple:

example

I think it should either color all of the non-glass boxes it hits (using e.g. `ray_hits_callback` to get many hits) or stop the laser at the first hit (draw it from left to the hit point with gizmos).

No need to squash the commits, I tend to squash merge like Bevy :)

Affinator:
OK, so the failed lints are from deprecated methods outside my pull request. Should I rebase on a different branch or can this be merged?

Jondolf:
I just pushed a quick fix for it here, it was simply a method that was renamed in a dependency. I think everything looks good, so I'll merge this once CI passes :)

PR #302, opened 2024-01-16
Add `CollisionPlugin` and support for generic colliders

Jondolf:
# Objective

Currently, collision detection is split into three separate plugins: `BroadPhasePlugin`, `NarrowPhasePlugin`, and `ContactReportingPlugin`. This modularity is great, but it would be convenient and user-friendly to have a higher level plugin that combined the functionality into one place.

In some cases, people might also want to use alternative backends for collision detection. For example, you could have SDF (Signed Distance Field) colliders, and would like them to work with the rest of the engine. This should ideally be as simple as passing the custom collider type to the `CollisionPlugin` through generics.

## Solution

- Add `CollisionPlugin` that adds collision detection plugins
- Add `AnyCollider` trait and use it in most plugins and systems that previously took `Collider`

Note: These changes are separate enough that I might split this PR in two.

---

## Todo

- [ ] Add custom collider example
- [ ] Add more docs

## Changelog

TODO

## Migration Guide

TODO


Jondolf:
Closing in favor of #311.

PR #310, opened 2024-01-23
fix raycast does not follow entity transform without rigidbody

### LINKED ISSUE #288: Raycast Does Not Follow Entity Transform Without Rigidbody
### Description:
We have identified a bug where the raycast functionality does not properly follow the transform of an entity if there is no Rigidbody component attached to the entity. Ideally, the raycast should track the entity's transform regardless of the presence of a Rigidbody.

### Expected Behavior:
The raycast should follow the transform of an entity, updating its direction and position based on the entity's current transform.

### Current Behavior:
When an entity does not have a Rigidbody component attached, the raycast remains static and does not update according to the entity's transform changes.

Steps to Reproduce:
1. Create an entity without a Rigidbody component.
2. Implement raycasting from this entity.
3. Alter the entity's transform (position, rotation, etc.).
4. Observe that the raycast does not follow the updated transform.

Workaround:
Currently, a temporary workaround is to attach a static Rigidbody to the entity. This can be done as follows:

```rust
commands.spawn((
    RayCaster::new(Vec3::ZERO, Vec3::NEG_Z),
    RigidBody::Static
));
```

zwazel:
# Objective

- Fixes #288 

## Solution

- If no Rotation and Position is available, fall back to GlobalTransform

## Changelog

- I've added a `From<GlobalTransform>` and `From<&GlobalTransform>` to `Position` as well as a `From<&GlobalTransform>` to `Rotation`. as a `From<GlobalTransform>` already existed.

I tested the updated `update_ray_caster_positions` system in my personal project, and it fixed the problem for me. I currently can't test the `update_shape_caster_positions` but as it is basically the same changes, i'd be surprised if it wouldn't work.

as this is my first time contributing to this crate, i'm not sure if the code is what is expected.

### DIFF
--- a/src/components/mod.rs
+++ b/src/components/mod.rs
@@ -331,6 +331,38 @@ impl Position {
     }
 }
 
+impl From<GlobalTransform> for Position {
+    #[cfg(feature = "2d")]
+    fn from(value: GlobalTransform) -> Self {
+        Self::from_xy(value.translation().x, value.translation().y)
+    }
+
+    #[cfg(feature = "3d")]
+    fn from(value: GlobalTransform) -> Self {
+        Self::from_xyz(
+            value.translation().x,
+            value.translation().y,
+            value.translation().z,
+        )
+    }
+}
+
+impl From<&GlobalTransform> for Position {
+    #[cfg(feature = "2d")]
+    fn from(value: &GlobalTransform) -> Self {
+        Self::from_xy(value.translation().x, value.translation().y)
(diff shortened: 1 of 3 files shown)


Jondolf:
Hi, sorry for leaving this sit here for so long! It seems to work though, thanks :)

I cleaned up the code a bit to remove the temp variables and make things more concise, and fixed some lint warnings that Clippy was giving

PR #312, opened 2024-01-24
Add more constructors for `SpatialQueryFilter`

Jondolf:
# Objective

`SpatialQueryFilter::new` is the same as `SpatialQueryFilter::default`, and constructing the filters can be a bit verbose. There should be more and better constructors.

## Solution

Change `SpatialQueryFilter::new` to take the masks and excluded entities, and add the `from_mask_bits`, `from_masks`, and `from_excluded_entities` constructors.

---

## Migration Guide

Change `SpatialQueryFilter::new()` to `SpatialQueryFilter::default()` or use the new constructors instead.


Jondolf:
Closing in favor of #313.

PR #336, opened 2024-02-23
Update example in Readme

RCoder01:
# Objective

The link in the current readme is broken and references an older version of the bevy 3d/3d_scene example.

## Solution

This PR updates the readme example to be based on the new bevy example and fixes the link.

There may be a better way to rotate the cylinder collider than using Collider::compound that I'm not aware of, feel free to update this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ fn setup(
 }
 ```
 
-<https://user-images.githubusercontent.com/57632562/230185604-b40441a2-48d8-4566-9b9e-be4825f4877e.mp4>
+<https://gist.github.com/assets/45217840/c096d786-06b9-4100-8f07-8d4c2c71f3c9>
 
 ## More examples
 
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ fn setup(
 }
 ```
 
-<https://gist.github.com/assets/45217840/c096d786-06b9-4100-8f07-8d4c2c71f3c9>
+<https://gist.github.com/assets/45217840/9a56edcf-03d5-4999-92fb-f577607b332c>
 
 ## More examples
 
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ bevy_xpbd_3d = "0.4"
 bevy_xpbd_3d = { git = "https://github.com/Jondolf/bevy_xpbd", branch = "main" }
 ```
 
-Below is a very simple example where a box with initial angular velocity falls onto a plane. This is a modified version of Bevy's [3d_scene](https://bevyengine.org/examples/3d/3d-scene/) example.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -95,55 +95,59 @@ fn main() {
...
-    // Plane
+    // circular base
     commands.spawn((
         RigidBody::Static,
-        Collider::cuboid(8.0, 0.002, 8.0),
+        Collider::compound(vec![(
+            Position(Vec3::ZERO),
+            Rotation(Quat::from_rotation_x(-std::f32::consts::FRAC_PI_2)),
+            Collider::cylinder(0.002, 4.0),
+        )]),
```
Jondolf, line 111: This is meant to be an extremely basic example that demonstrates the basic API, so this shouldn't be using a compound collider and `Position` and `Rotation` here. It's needed since you rotate the circle mesh, which also rotates this cylinder, but it makes the example way too confusing for an introductory example.

I'd just use the old cuboid/plane here, for both the mesh and collider, since it's a lot simpler. As mentioned above this example, this is a modified version of Bevy's example, so we can change some things. We could also just remove the link to the example completely and have this be a standalone thing

README.md
```diff
@@ -95,55 +95,59 @@ fn main() {
...
-    // Cube
+    // cube
     commands.spawn((
         RigidBody::Dynamic,
         AngularVelocity(Vec3::new(2.5, 3.4, 1.6)),
         Collider::cuboid(1.0, 1.0, 1.0),
         PbrBundle {
-            mesh: meshes.add(Cuboid::default()),
-            material: materials.add(Color::rgb(0.8, 0.7, 0.6)),
+            mesh: meshes.add(Cuboid::new(1.0, 1.0, 1.0)),
```
Jondolf, line 127: There's no need to change this imo

```suggestion
            mesh: meshes.add(Cuboid::default()),
```

README.md
```diff
@@ -95,55 +95,59 @@ fn main() {
...
+            material: materials.add(Color::rgb_u8(124, 144, 255)),
             transform: Transform::from_xyz(0.0, 4.0, 0.0),
             ..default()
         },
     ));
-
-    // Light
+    // light
```
Jondolf, line 133: Nit, but I prefer to start comments with a capital letter, and have the extra newline between spawns, like what it was before. Bevy's *[own style guide](https://github.com/bevyengine/bevy/blob/main/.github/contributing/engine_style_guide.md)* says that comments should be capitalized, but it's contradicting itself in this example

```suggestion

    // Light
```

(this would of course apply to all other comments here too)

Jondolf:
Hi, I added my own version of this example along with an updated video/gif in f231f86. Closing in favor of that :)

RCoder01:
Sorry I forgot about this PR since I opened it, thanks for updating the example. I noticed the link was still broken though: #396.

PR #344, opened 2024-03-01
Fix 3D ShapeCaster global_rotation

ramon-oliveira:
# Objective

Fix 3D `ShapeCaster` global rotation. Currently, shape and parent rotation are adding its `Quat` instead of multiplying.

## Solution

Multiply the rotations instead of adding.

-----

## Before and After

https://github.com/Jondolf/bevy_xpbd/assets/4665205/7aa92bec-fa62-4810-b518-945fd9896859


Jondolf (approved):
Good fix, thanks!

PR #345, opened 2024-03-02
Normalize rotations after solving constraints in solver

### LINKED ISSUE #235: rotations not normalized in solver, causing colliders to expand and eventually explode

Jondolf:
# Objective

Adopted from #237, fixes #235.

Constraints update rotation by adding or subtracting quaternions. Unlike multiplying unit quaternions (the typical way to "add" together rotations), this can result in rotations becoming unnormalized. This can cause explosions (see #235), and cause panics when trying to rotate Bevy's direction types.

## Solution

Normalize the rotations after the constraint rotation delta has been applied. This does come with a small extra cost, but I believe it is worth it for stability. The normalization is also present in some other XPBD implementations I've found, like [this one](https://github.com/felipeek/raw-physics/blob/96fbc46eb8f89d007967c219d2c22e74d7375fd1/src/physics/pbd_base_constraints.cpp#L111).


mattdm:
Sorry I didn't get back to this! January was busier than anticipated, and I've had covid for the whole last month. I'll be around again sometime!

PR #347, opened 2024-03-08
Use `compile_error!` macro instead of panicking in `PhysicsLayer` derive macro

doonv:
# Objective

Improve errors in the `PhysicsLayer` derive macro

## Solution

Instead of panicking, we can instead return a `compile_error!` macro. This removes the "proc-macro panicked" message and also allows us to specify a span (location in the code) for the error.

### Future solutions

I'm thinking of also adding a `nightly` feature which uses the unstable diagnostic API for even better errors.

---

## Changelog

- Improved errors in the `PhysicsLayer` derive macro


Jondolf (approved):
Looks great, thanks!

PR #354, opened 2024-03-20
Fix some doc tests

yrns:
Default features failed on these. I guess CI doesn't do the `debug-plugin` or `f32` features so it didn't catch them.


Jondolf (approved):
Thanks!

PR #355, opened 2024-03-29
Sensors and mass properties

yrns:
# Objective

I was surprised to find that a sensor added as a child of a dynamic body affected the parent's behavior. Even with the density set to zero the difference was noticeable.

## Solution

- Make the default mass and density zero for sensors.

and/or:

- Ignore sensors when updating parents' mass properties.

Maybe instead of adding a special case for sensors (which would be a breaking change), a note should be added to the sensor documentation indicating you need to set the mass properties to zero if you don't want to affect any parents?

---

## Changelog

tbd

## Migration Guide

tbd

### DIFF
--- a/src/plugins/collision/collider_backend.rs
+++ b/src/plugins/collision/collider_backend.rs
@@ -238,16 +238,32 @@ fn init_colliders<C: AnyCollider>(
             Option<&ColliderAabb>,
             Option<&ColliderDensity>,
             Option<&ColliderMassProperties>,
+            Has<Sensor>,
         ),
         Added<C>,
     >,
 ) {
-    for (entity, collider, aabb, density, mass_properties) in &mut colliders {
-        let density = *density.unwrap_or(&ColliderDensity::default());
+    for (entity, collider, aabb, density, mass_properties, is_sensor) in &mut colliders {
+        // Default zero density and mass for sensors.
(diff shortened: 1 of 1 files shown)


Jondolf:
Hi, I do agree we should change this. I looked into it a bit, and it seems like the sensor equivalents in Godot (Area2D/Area3D) and Unity (trigger property) also don't contribute to the mass properties of the rigid body.

I think the semantics we should go for are that sensors don't affect physics *at all*, and are only for detecting intersections. The way I would do this is slightly different from this PR though:

- Don't run `update_collider_mass_properties` for sensors. This PR already does this :)
- When `Sensor` is added to a collider, remove the collider's mass properties (if any) from the rigid body's mass properties.
- When `Sensor` is removed from a collider, update the collider's mass properties and add them to the rigid body's mass properties.

I think this could be implemented robustly in Bevy 0.14 using component lifecycle hooks. Notably different from this PR, I would not make the default `ColliderDensity` zero, because it is rather implicit, and if I later removed the `Sensor` component from the collider, I would expect that it would have mass properties again like a default collider.

Do you mind if I adopt this and implement it for the upcoming Bevy 0.14 update with lifecycle hooks? I can still add you as a co-author :)

yrns:
All that sounds good. Thanks for taking a look.

Jondolf:
Closing in favor of #381 :)

PR #364, opened 2024-05-17
add `IgnoredCollisions` component

exoexo-dev:
# Objective

Allow to set ignored collisions per entity.

## Solution

Added an `IgnoredCollisions` component which is checked in the broad phase to filter out ignored collisions before any detection checks.

---

### Added

Added an `IgnoredCollisions` component which can be added to entities with a `Collider` to configure a set of entities with which there will be no collision detection with the owning entity.

### DIFF
--- a/src/components/layers.rs
+++ b/src/components/layers.rs
@@ -237,6 +237,8 @@ impl Not for LayerMask {
 /// Colliders without this component can be considered as having all memberships and filters, and they can
 /// interact with everything that belongs on any layer.
 ///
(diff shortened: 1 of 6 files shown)


exoexo-dev:
Looks like something is still missing here - when the entities with ignored collisions are overlapping on spawn they still "explode"

exoexo-dev:
> Marking as Waiting-on-Author due to merge conflicts

Hi, I rebased but there is some weird linker error in CI tests now

PR #367, opened 2024-06-10
Implement `RegularPolygon` colliders with a custom shape

Jondolf:
# Objective

Currently, regular polygons don't have their own collider shape. Instead, the `From<RegularPolygon>` implementation computes the convex hull and creates a convex polygon from that. This may not be the most efficient approach however, as some computations can be simpler and more efficient if the polygon is known to be regular.

This PR adds a custom shape implementation for the 2D `RegularPolygon` primitive.

## Solution

Add a `RegularPolygonWrapper` type that is used as the internal representation for the custom regular polygon shape. The custom shape uses the ID `2`.

---

## Migration Guide

If you had a custom Parry shape using the ID `2`, you might need to change it, because the ID is used for regular polygons.

### DIFF
--- a/src/math/double.rs
+++ b/src/math/double.rs
@@ -3,8 +3,12 @@ use bevy_math::*;
 
 /// The floating point number type used by Bevy XPBD.
 pub type Scalar = f64;
+/// The PI/2 constant.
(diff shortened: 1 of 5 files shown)


Nul-led:
lgtm

PR #372, opened 2024-06-16
Update bevy-0.14 branch to crates.io release of nalgebra 

gmorenz:
Glam 0.27 support was [merged and released in 0.3.26](https://github.com/dimforge/nalgebra/pull/1390) so I think this should work with the crates.io version now. 

The tests don't all pass, but that doesn't look like it's related to nalgebra, just the bevy update.


Jondolf (approved):
Thanks! Seems to work trying this locally.

PR #377, opened 2024-06-19
Speed up `ColliderTransform` propagation and extract collider hierarchy logic into `ColliderHierarchyPlugin`

### LINKED ISSUE #356: Awful performance with many non-physics entities
We had discussed this on Discord previously, but it seems like the issue still persists in the latest version.

We have a 2D game where we use LDTK (via `bevy_ecs_ldtk`, which is based on `bevy_ecs_tilemap`). When the level is spawned, that creates a lot of ECS entities in Bevy (>60k for a moderately sized level). None of those entities have any physics components, they are
just to represent the visuals of the map.

We have very few physics entities (<100 colliders for walls in the level and a few kinematic bodies, no dynamic bodies yet).

The game runs awfully slowly . I ran a trace to determine what is going on, and it looks like `bevy_xpbd_2d`'s `propagate_collider_transforms` system is the culprit.

That system takes 550-600 us every sub-tick, adding up to 8ms for the whole PhysicsSchedule.

It seems like it is iterating through all of those non-physics entities and wasting CPU and time.

In conclusion: `bevy_xpbd` scales very poorly for applications with few physics entities and lots of non-physics entities. A lot of work is wasted on non-physics entities to process their "collider transforms" (even though they do not have colliders or any other physics components).

Jondolf:
# Objective

**Note:** This is for the `bevy-0.14` branch, because it uses new observer functionality (although it's not strictly necessary). It will be merged to `main` for the upcoming release.

Fixes #356.

To handle transforms in collider hierarchies properly, the engine must currently propagate a `ColliderTransform` component down the hierarchy several times per frame. This is made even worse by the fact that this propagation is done at every *substep*, not just once or twice per frame.

This wouldn't be too bad if it only affected physics entities, but the current system traverses through *all* entities that have a `Transform`. As reported in #356, this leads to absolutely awful performance for scenes with a large number of non-physics entities. This means that the engine scales very poorly with entity count, which is unacceptable.

### Speed up propagation with `ColliderAncestor`

When colliders are added as children, automatically mark all ancestors with the `ColliderAncestor` component. Similarly, clean up the markers when a collider is removed.

`ColliderTransform` is *only* propagated down trees with `ColliderAncestor` entities. This way, we can essentially skip all unnecessary propagation.

In a test scene with one root entity, 100,000 child entities, and 12 substeps, I was previously getting **~22 fps** even though the entities only had a `SpatialBundle` and no physics components. With the changes in this PR, I now get **~200 fps**. Adding colliders to individual entities doesn't impact performance in any meaningful way, as the propagation is only done for those entities.

Note that this propagation will be even less of an issue in the upcoming Avian release, because there it doesn't even need to be run in the substepping loop.

The `ColliderAncestor` logic has a pretty comprehensive unit test for adding and removing colliders in a hierarchy, but I haven't verified yet if simply changing the `Parent` without moving the `Collider` works as expected. Adding more tests and potentially fixing these more niche cases can be done in a follow-up.

### Add `ColliderHierarchyPlugin`

The `ColliderBackendPlugin` was getting *big*, so I decided to extract the logic related to hierarchies and `ColliderTransform` into a new `ColliderHierarchyPlugin`.

Ideally, you could use the plugins fully independently and for example completely abandon hierarchies if you wanted to, but currently, they are still lightly coupled. Eliminating this coupling is a task for a follow-up.

---

## Migration Guide

Hierarchy and transform logic for colliders has been extracted from the `ColliderBackendPlugin` into a new `ColliderHierarchyPlugin`, which by default is included in the `PhysicsPlugins` plugin group.

If you are adding plugins manually, make sure you have both if you want that functionality.


Jondolf:
I have the follow-up implemented, and it reworks this further. I'll merge this first though, and open the follow-up right after

PR #381, opened 2024-06-22
Make sensors not contribute to mass properties

Jondolf:
# Objective

Similar to #355 by @yrns, but without special casing `ColliderDensity` defaults, and implemented with observers.

Sensor colliders currently contribute to the mass properties of rigid bodies. This differs from most peoples' expectations, and is also different from many existing engines such as Godot and Unity. Sensor colliders should have no impact on the physics simulation.

## Solution

- When the `Sensor` component is added to an entity, remove the collider's contribution on the rigid body's mass properties.
- When the `Sensor` component is removed from an entity, add the collider's mass properties to the rigid body's mass properties.

## Testing

There is a test to make sure that the mass properties of a rigid body are updated accordingly when the `Sensor` component is first added and then removed from a collider.

---

## Migration Guide

Colliders with the `Sensor` component no longer contribute to the mass properties of rigid bodies. You can add mass for them by adding another collider that is *not* a sensor, or by manually adding mass properties with the `MassPropertiesBundle` or its components.

Additionally, the mass properties of `Sensor` colliders are no longer updated automatically, unless the `Sensor` component is removed.

### DIFF
--- a/src/plugins/collision/collider/backend.rs
+++ b/src/plugins/collision/collider/backend.rs
@@ -131,6 +131,74 @@ impl<C: ScalableCollider> Plugin for ColliderBackendPlugin<C> {
                 );
             });
 
+        // When the `Sensor` component is removed from a collider,
+        // remove the collider's contribution on the rigid body's mass properties.
+        app.observe(
+            |trigger: Trigger<OnAdd, Sensor>,
+             query: Query<(
+                &ColliderParent,
+                &ColliderMassProperties,
+                &PreviousColliderTransform,
+            )>,
+             mut body_query: Query<MassPropertiesQuery>| {
+                if let Ok((
+                    collider_parent,
+                    collider_mass_properties,
+                    previous_collider_transform,
+                )) = query.get(trigger.entity())
+                {
+                    if let Ok(mut mass_properties) = body_query.get_mut(collider_parent.0) {
+                        // Subtract previous collider mass props from the body's own mass props.
+                        mass_properties -= ColliderMassProperties {
+                            center_of_mass: CenterOfMass(
+                                previous_collider_transform
+                                    .transform_point(collider_mass_properties.center_of_mass.0),
+                            ),
+                            ..*collider_mass_properties
+                        };
+                    }
+                }
+            },
+        );
+
+        // When the `Sensor` component is added to a collider,
+        // add the collider's mass properties to the rigid body's mass properties.
+        app.observe(
+            |trigger: Trigger<OnRemove, Sensor>,
+             mut collider_query: Query<(
+                Ref<C>,
+                &ColliderParent,
+                &ColliderDensity,
+                &mut ColliderMassProperties,
+                &ColliderTransform,
+            )>,
+             mut body_query: Query<MassPropertiesQuery>| {
+                if let Ok((
+                    collider,
+                    collider_parent,
+                    density,
+                    mut collider_mass_properties,
+                    collider_transform,
+                )) = collider_query.get_mut(trigger.entity())
+                {
+                    if let Ok(mut mass_properties) = body_query.get_mut(collider_parent.0) {
+                        // Update collider mass props.
+                        *collider_mass_properties =
+                            collider.mass_properties(density.max(Scalar::EPSILON));
+
+                        // Add new collider mass props to the body's mass props.
+                        mass_properties += ColliderMassProperties {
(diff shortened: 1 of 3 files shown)


src/plugins/collision/collider/mod.rs
```diff
@@ -420,7 +422,13 @@ impl From<Transform> for ColliderTransform {
 /// fn setup(mut commands: Commands) {
 ///     // Spawn a static body with a sensor collider.
 ///     // Other bodies will pass through, but it will still send collision events.
-///     commands.spawn((RigidBody::Static, Collider::ball(0.5), Sensor));
+///     let collider = Collider::ball(0.5);
+///     commands.spawn((
+///         RigidBody::Static,
+///         MassPropertiesBundle::new_computed(&collider, 1.0),
```
janhohenheim, line 428: Why does this sensor have a `MassPropertiesBundle`?

src/plugins/collision/collider/backend.rs
```diff
@@ -131,6 +131,74 @@ impl<C: ScalableCollider> Plugin for ColliderBackendPlugin<C> {
...
+             mut body_query: Query<MassPropertiesQuery>| {
+                if let Ok((
+                    collider_parent,
+                    collider_mass_properties,
+                    previous_collider_transform,
+                )) = query.get(trigger.entity())
+                {
+                    if let Ok(mut mass_properties) = body_query.get_mut(collider_parent.0) {
+                        // Subtract previous collider mass props from the body's own mass props.
+                        mass_properties -= ColliderMassProperties {
```
janhohenheim, line 152: Am I understanding the following right?
- When a `Collider` is added, its `ColliderParent` gets mass properties added
- When a `Sensor` is added, its `ColliderParent` gets mass properties removed

Ideally, these two then cancel each other out.
Assuming I got the above correct:
Is the ordering relevant? Can the sensor part be executed first, resulting in a temporary negative collider mass property?

src/plugins/collision/collider/backend.rs
```diff
@@ -131,6 +131,74 @@ impl<C: ScalableCollider> Plugin for ColliderBackendPlugin<C> {
...
+                    collider_transform,
+                )) = collider_query.get_mut(trigger.entity())
+                {
+                    if let Ok(mut mass_properties) = body_query.get_mut(collider_parent.0) {
+                        // Update collider mass props.
+                        *collider_mass_properties =
+                            collider.mass_properties(density.max(Scalar::EPSILON));
+
+                        // Add new collider mass props to the body's mass props.
+                        mass_properties += ColliderMassProperties {
```
janhohenheim, line 190: These lines are duplicated. I recommend adding the following function:
```rust
impl ColliderMassProperties {
  fn with_collider_transform(self, transform: &ColliderTransform) -> Self {
    Self {
      center_of_mass: CenterOfMass(
        collider_transform.transform_point(self.center_of_mass.0),
      ),
      ..self
    }
  }
}
```

janhohenheim (changes requested):

src/plugins/collision/collider/backend.rs
Jondolf: If you added `Sensor` first, the entity probably wouldn't have `ColliderParent` or `ColliderMassProperties`, since those are only added automatically for colliders.

If the collider and `Sensor` were added at the same time, with the current setup, the collider wouldn't have mass properties computed for it yet because the observer should run before `init_colliders`. So I believe it shouldn't be an issue currently.

However, if we later changed `init_colliders` to use an observer or lifecycle hook as well, the order would be more ambiguous, which could cause issues. A potential solution could be to initialize sensors with zero mass properties (we already don't run `update_collider_mass_properties` for them anyway) and only subtract from the rigid body's mass properties in the observer if the collider mass properties are non-zero.

src/plugins/collision/collider/mod.rs
Jondolf: Oops, somehow I didn't notice that it's static and just thought "oh no, this would be a rigid body with zero mass!" which would only be problematic if this was dynamic 😅 Fixing

src/plugins/collision/collider/backend.rs
Jondolf: Good idea! I named it `transformed_by` though to be more similar to Bevy's transformation methods on e.g. `Mesh` and bounding volumes

janhohenheim (approved):

Issue #346, opened 2024-03-05
Rebrand Bevy XPBD and Switch Solvers


Jondolf:
I am making this as an early notice of major upcoming changes and to explain their rationale. There are important issues that need to be addressed, and I want to be as transparent as possible.

**Within the next couple of months, we will attempt to transition away from *Extended Position-Based Dynamics* (XPBD) and rebrand `bevy_xpbd`. This involves a partial rewrite of the internals and a rename for the crate.**

There is a lot of text here, so if you want to skip to how this might affect you, feel free to skip ahead to the [Questions](#Questions) and [Conclusion](#Conclusion) sections. For full context, however, I recommend reading on.

## Why?

`bevy_xpbd` started as a small project with the idea of trying to build an ECS-first physics engine for Bevy using this shiny new simulation method called *Extended Position-Based Dynamics*. XPBD looked very promising and intriguing, providing great stability through an innovative substepping scheme and expressive position-based constraints.

However, as I've come to learn over the crate's development, using XPBD comes with several issues. Perhaps the most serious one is that XPBD is technically [patented](https://uspto.report/patent/grant/11,270,041) by NVIDIA.

There is no prior record of NVIDIA coming after projects using XPBD, and there are several existing open source implementations of it, but the patent still exists. This is a potential risk to me as the author of `bevy_xpbd`, people distributing projects using it, and to the ecosystem as a whole. `bevy_xpbd` can *not* be officially adopted by Bevy as long as it uses XPBD, nor can it be recommended by its maintainers. For the benefit of everyone involved, we must transition away from XPBD.

As an initial action, I have added a warning about this in the README that links to this issue. For projects using `bevy_xpbd` right now, there should be no reason to panic, but it is important to be aware of the legal concerns. While this is very unlikely to cause issues for smaller projects in practice, it is technically up to the user to have the rights to distribute the library in their works.

Aside from potential patenting issues, XPBD also has some problems on the simulation side of things.

- Overlap is solved by pushing bodies apart at a position-level. This makes cases where bodies are constantly overlapping (like squished in a confined box) very energetic and potentially explosive.
- Position-based constraints can often have high-frequency oscillation where bodies are always slightly moving back-and-forth. Looking up close, this can sometimes be noticeable. XPBD rarely (if ever) reaches a completely stable and relaxed state without sleeping.
- Friction is typically not quite as robust as it is with impulse-based methods. There are several ways of handling it, some better than others, but it can still have some issues.
[...]

PR #385, opened 2024-06-25
Rework contact solver and collision detection, implement speculative collision

Jondolf:
# Objective

*Note: This is for the `avian` branch, which will later be pushed to `main`.*

Bevy XPBD currently uses Extended Position-Based Dynamics for its solver. While it has worked quite well, it does have several issues:

- Solving deep overlap is very energetic and explosive.
- Position-based constraints often have high-frequency oscillation where bodies are always slightly moving back-and-forth. XPBD rarely (if ever) reaches a truly stable and relaxed state without sleeping.
- A ton of substeps are needed for minimizing jitter and making systems of bodies stable.
- Friction is typically not quite as robust as it is with impulse-based methods. There are several ways of handling it, some better than others, but it can still have some issues.
- Many features can be more challenging to implement and expose APIs for, and there are fewer learning resources and references for the usage of XPBD in game physics.

Additionally, XPBD even has some potential legal ambiguities, see #346.

Aside from XPBD itself, there are also several other issues with the current collisions and rigid body dynamics:

- Narrow phase collision detection is run at every substep instead of just once per frame like most physics engines. This can be very bad for performance.
- There is no support for any kind of Continuous Collision Detection, so tunneling can be a big issue.
- It is not possible to configure the behavior of collisions aside from the number of substeps.
- #224
- Sleeping in 0.4 just doesn't work.

These are big issues, and I don't currently see XPBD as the best option for the contact solver going forward. Extensive rewrites and fixes are needed.

## Solution

This PR consists of three major parts:

- Rewrite the contact solver to use an impulse-based TGS Soft solver instead of XPBD.
- Rework the narrow phase and other collision detection logic.
- Implement speculative collision, a form of Continuous Collision Detection.

They all relate to each other, and are covered in detail below.

### Substepped Impulse-Based Solver With Soft Constraints

The contact solver has been rewritten to use **TGS Soft**, a substepped impulse-based solver using soft constraints. The choice was largely motivated by the wonderful Erin Catto's [Solver2D] experiments, where TGS Soft was deemed as quite a clear winner.

Box2D V3 was used as the primary inspiration for the core implementation of the new solver. Engines such as Rapier and Bepu also use a very similar approach.

#### Terminology

- **Projected Gauss-Seidel (PGS)**: The classic iterative approach to solving constraints (contacts and joints) using the [Gauss-Seidel](https://en.wikipedia.org/wiki/Gauss%E2%80%93Seidel_method) numerical method. Reframed by Erin Catto as [Sequential Impulses](https://box2d.org/files/ErinCatto_SequentialImpulses_GDC2006.pdf).
- **Temporal Gauss-Seidel (TGS)**: Same as PGS, but prefers *substepping* over *iteration*, running more simulation steps with smaller time steps rather than simply iteratively solving constraints. Substeps tend to be more effective than iterations, as shown in the [Small Steps in Physics Simulation](https://mmacklin.com/smallsteps.pdf) paper by Macklin et al.
- **Baumgarte stabilization**: When solving contact constraints using impulses, boost the impulses using a **bias** to account for overlap and actually push the bodies apart.
- **Soft constraints**: Similar to Baumgarte stabilization, but more stable and controlled. Based on the [harmonic oscillator](https://en.wikipedia.org/wiki/Harmonic_oscillator), soft constraints dampen constraint responses, and can be tuned intuitively with a frequency and damping ratio.
- **Warm starting**: Store the constraint impulses from the previous frame and initialize the solver by applying them at the start of the current frame. This helps the solver converge on the solution faster, and is especially helpful when objects are coming to rest.
- **Relaxation**: Baumgarte stabilization and soft constraints can add unwanted energy. Relaxation helps remove it by solving constraints a second time, but without a bias.

Please refer to the [Solver2D] post for a more complete overview of what TGS and soft constraints are, how they work, and how they relate to other approaches.

#### Solver Overview

As stated earlier, contacts use TGS Soft. However, joints still currently use XPBD, so the new solver is actually a kind of hybrid solver. I do plan on transitioning joints away from XPBD in the future though.

Below is a high-level overview of the new structure of the solver.

1. Broad phase collision detection collects potential collision pairs into `BroadCollisionPairs`.
2. Narrow phase collision detection computes contacts for the pairs and adds them to `Collisions`.
3. A `ContactConstraint` is generated for each contact manifold, and added to `ContactConstraints`.
4. Substepping loop, running `SubstepCount` times.
	1. Integrate velocities, applying gravity and external forces.
	2. Warm start the solver.
	3. Solve velocity constraints with bias (soft constraints).
	4. Integrate positions, moving bodies based on their velocities.
	5. Relax velocities by solving constraints again, but *without* bias.
	6. Solve XPBD constraints (joints) and perform XPBD velocity projection.
5. Apply restitution as a post-phase.
6. Finalize positions by applying `AccumulatedTranslation`.
7. Store contact impulses for next frame's warm starting.

Refer to the code for implementation details. The contact logic and constraints are quite heavily commented and should hopefully be relatively easy to follow.

#### New Solver Results

Collisions have significantly less drift than before, and performance is much better. Below is a pyramid with a base of 64 boxes, simulated with 4 substeps, with sleeping disabled.

**Old**: XPBD has substantial drift, and the pyramid quickly collapses in on itself. This even happens with a much larger number of substeps, although to a slightly lesser degree. Performance is very poor, even with just 4 substeps. 

https://github.com/Jondolf/bevy_xpbd/assets/57632562/ca0ff2f5-dfad-4395-a662-be41cdd7bfcf

**New**: With TGS Soft, the pyramid stays stable. There is a very small amount of drift over a long period of time, but even that can be mitigated by configuring the contact behavior through the `SolverConfig` and/or by adding more substeps.

https://github.com/Jondolf/bevy_xpbd/assets/57632562/a4a48e95-d8e9-459e-bfbb-26d0d12bb8ac

Impulses even out and stabilize *much* better with TGS Soft, even with deep overlap. In overlapping cases, the old XPBD implementation was significantly more explosive. Below is an example where colliders are dynamically enlarged to fill up a container.

**Old**: With XPBD, overlap causes explosions even before the shapes fill the container. Once they do fill the container, they jitter very strongly and explode through the walls. This is typically very undesirable for games.

https://github.com/Jondolf/bevy_xpbd/assets/57632562/e2d47c7e-b03a-45a7-93f9-8c5909564af0

**New**: Overlap is solved perfectly smoothly with no explosions. The contact impulses even out without jitter. With no space to go, the shapes prioritize stability over perfectly hard contacts that would cause explosiveness.

https://github.com/Jondolf/bevy_xpbd/assets/57632562/78f3d158-42a1-4841-b68d-3a6e3e06e451

An important thing to note is that the new solver uses several tolerances and thresholds that are length-based. The old solver also had some tolerances, but now they are much more important for stability.

Without any tuning, a 2D game using pixels as length units might have collision issues, because the tolerances would be wrong for that application. For example, below is a scene with stacks of balls that have a radius of 5.0, with no tuning whatsoever:

No tuning

The contacts are too soft. To fix this, there is a new `PhysicsLengthUnit` resource, which can be thought of as a kind of pixels-per-meter conversion factor. It is only for scaling the internal tolerances (and debug rendering gizmos!) however, and doesn't scale objects or velocities in any way.

`PhysicsLengthUnit` can be easily set when adding `PhysicsPlugins` for an app:

```rust
fn main() {
    App::new()
        .add_plugins((
            DefaultPlugins,
            // A 2D game with 20 pixels per meter
            PhysicsPlugins::default().with_length_unit(20.0),
        ))
        .run();
}
```

And with that, we get the stability and behavior we expect:

With the appropriate length unit

[Solver2D]: https://box2d.org/posts/2024/02/solver2d/

### Collision Detection Refactor

The narrow phase and the `PostProcessCollisions` schedule are now run in the `PhysicsSchedule` instead of the `SubstepSchedule`. This greatly improves performance, and also makes collision events and reacting to collisions less awkward and footgunny: previously, the contact data was always from the *last* substep, but at that point, the contact is often already mostly resolved.

Moving narrow phase collision detection out of the substepping loop is possible by simply storing contact data in local space and computing the updated separation distance at each substep using the current poses of the bodies. This way, the contact data is (approximately) accurate even with the bodies moving relative to each other within substeps.

A few other improvements have been made as well:

- The narrow phase logic has been extracted into a `NarrowPhase` system parameter.
- The broad phase now outputs intersection pairs with the entities in ascending order instead of based on the minimum X extent. This fixes issues where bodies moving past each other on the X axis counts as a separate collision, causing problems with collision events.
- #224 is fixed.
- The warnings logged for overlapping bodies have less false alarms.

### Speculative Collision

Tunneling is a phenomenon where fast-moving small objects can pass through thin geometry such as walls due to collision detection being run at discrete time steps:

Tunneling

Moving the narrow phase out of the substepping loop has the unfortunate consequence that it increases the risk of tunneling as collisions are not computed as frequently. A solution for this is needed before we commit to the narrow phase change.

One of the primary solutions to tunneling is **Continuous Collision Detection**. There are two common forms:

1. **Sweep-based CCD**: Each collider with CCD enabled is swept from the current position to the predicted one (or from the previous position to the current one). If a hit is detected, the bodies are moved to the time of impact. Contact resolution can be left to the next frame, or to avoid "time loss", performed immediately with a substepping scheme that can also consider secondary collisions.
2. **Speculative CCD**: Contact points are predicted by the narrow phase for fast-moving objects before they actually touch. Speculative collision response is handled by pushing back the part of velocity that would cause penetration.

Sweep-based CCD can be much more expensive and more complex, especially with substepping and non-linear sweeps. Speculative collision on the other hand is very cheap while still being quite robust, although in rare cases it can miss collisions or cause ghost collisions.

The new solver implements speculative collision, which is enabled for *all* bodies *by default*. In my experiments, this is efficient, improves stability, and eliminates almost *all* tunneling except when caused by contact softness. This approach seems to also be taken by both Bepu and Box2D.

Below is a high-level overview of how speculative collision is currently implemented.

- The speculative margin is the maximum distance at which a collision pair generates speculative contacts. This is unbounded for every rigid body by default, which eliminates almost all tunneling.
- Each AABB is expanded in the movement direction based on the body's velocity, clamped by the speculative margin if it is bounded. (Box2D might have a way to avoid this expansion, but I haven't looked into it yet)
- The effective speculative margin is the actual margin used for contact computation, and it is clamped based on the velocities. This is used as the maximum separation distance for contact / closest point computation.
- When actually solving the contact (normal part, no friction), use the softness parameters only if the contact is penetrating. Otherwise it is speculative, and we bias the impulse to cancel out the velocity that would cause penetration.
- Apply restitution in a separate phase after the substepping loop.

Speculative collision

From a user's point of view, this happens completely in the background. However, if desired, the speculative margin can be configured for an entity using the `SpeculativeMargin` component, or even globally using `default_speculative_margin` in `NarrowPhaseConfig`.

Below is an example of balls being shot at high speeds at thin walls and spinning objects, with just a single substep to eliminate the effect of substepping. The simulation is paused and stepped manually a few times to closer see the behavior.

**Old**: Almost all of the balls simply pass through the thin geometry. With more substeps, this tunneling could been reduced, but never removed completely. Increasing the substep count would also hurt performance.

https://github.com/Jondolf/bevy_xpbd/assets/57632562/03f9f623-0c5f-4228-a913-29bfaad23dc4

**New**: The balls never tunnel straight through the walls, and often even hit the spinning objects. You can still see many balls phasing through the walls, but this is mostly due to contact softness, in cases where a body hits another body hard enough to force it through the wall.

https://github.com/Jondolf/bevy_xpbd/assets/57632562/70526783-762a-452c-87fd-fe23789560f9

Of course, bodies getting pushed through the ground by other bodies is still an issue. This could be reduced in the future by solving contacts against static objects *after* dynamic-dynamic contacts, giving them higher priority and stiffness. This is follow-up material however, and cases where this is an issue should be quite rare in games.

Sweep-based CCD will also be added as an option in a follow-up. I already have an implementation ready locally.

## Performance Results

Comparing this branch to the `avian` branch (which still has the old solver), there is a roughly 4-5x performance improvement for collision-heavy scenes, with the difference growing with the number of collisions and the number of substeps.

The benchmarks below use 8 substeps.

Benchmark with 8 substep

With just a single substep, the difference is smaller, but this branch is still faster, up to 2x.

Benchmark with 1 substep

## Other Changes

This is such a big rework that it is unfortunately almost impossible to cover every change. Below are some noteworthy changes however.

### Sleeping Rework

- Sleeping in 0.4 was broken, and bodies never fell asleep properly. This has been fixed.
- Bodies now store a `PhysicsChangeTicks` component. This is used to detect what component changes are made by the user *outside* of physics schedules, and what changes are made by the physics engine. This way, we can ignore changes made by the physics engine and have more control over when bodies are woken up.
- For now, bodies can only sleep when they are not in contact with other dynamic bodies. This is because the current per-body sleeping approach is quite buggy and has stability issues. In the future, I plan on implementing simulation islands and doing per-island sleeping like most other physics engines.

### Integrator Rework

- Velocity integration and position integration now run in separate systems and in different parts of the schedule: `IntegrationSet::Velocity` and `IntegrationSet::Position`. This is needed for the new solver.
- The semi-implicit Euler integration scheme now has its own module with proper docs and tests.
- Integration uses `par_iter_mut`.

### Scheduling Changes

- `PhysicsStepSet` now has `First` and `Last` variants, so users can easily schedule systems before or after physics in the `PhysicsSchedule`.
- The narrow phase and the `PostProcessCollisions` schedule are now run in `PhysicsStepSet::NarrowPhase` instead of `SubstepSet::NarrowPhase`.
- Integration is now run in `IntegrationSet::Velocity` and `IntegrationSet::Position` instead of `SubstepSet::Integrate`.
- `SubstepSet` has been removed.
	- The solver runs in `PhysicsStepSet::Solver`.
	- The solver's system sets are in `SolverSet`.
	- Substepping is performed in `SolverSet::Substep`.
	- The substepping loop's system sets are in `SubstepSolverSet`.

### New Configuration Options

- `NarrowPhaseConfig` has new `default_speculative_margin` and `contact_tolerance` properties.
- The new `SolverConfig` resource can be used for tuning collisions.
- The new `PhysicsLengthUnit` resource can be used as a scaling factor for internal length-based tolrances.
	- Existing length-based tolerances and thresholds are now scaled by this, like the `linear` property of `SleepThreshold`.
	- Debug rendering is also scaled by the length unit.

### Examples

- `XpbdExamplePlugin` has been renamed to `ExampleCommonPlugin`, and it no longer adds `PhysicsPlugins` automatically.
- 2D examples have `PhysicsLengthUnit` configured.
- The `one_way_platform` example's logic has been modified to account for `PostProcessCollisions` no longer running in the `SubstepSchedule`.
- The collision logic for the kinematic character controller examples has been rewritten to run in the `PhysicsSchedule` without jitter or stability issues.

### Miscallaneous

- `ColliderAabb` now has the `grow` and `shrink` methods.
- `ContactData` now stores feature IDs for contact matching, which is needed for warm starting.
- `ContactData` property `index` has been removed.
- 3D tangent impulses are now 2D vectors instead of scalar values.

---

### New Contact Solver

The contact solver has been rewritten. In practice, this has the following effects:

- Collisions should be much more stable.
- Resolving overlap is no longer nearly as explosive.
- Less substeps are generally needed for stability.
- Tunneling is much more rare.
- Performance is better.

However:

- Contacts may even be *too* soft by default for some applications. This can be tuned with the `SolverConfig`.
- Static friction is currently not considered separately from dynamic friction. This may be fixed in the future.
- Restitution might not be quite as perfect in some instances (this is a tradeoff for speculative collision to avoid tunneling).
- 2D applications may need to configure the `PhysicsLengthUnit` to get the best stability and behavior.

The `PhysicsLengthUnit` can be thought of a pixels-per-meter scaling factor for the engine's internal length-based tolerances and thresholds, such as the maximum speed at which overlap is resolved, or the speed threshold for allowing bodies to sleep. It does *not* scale actual physics objects or their velocities.

To configure the `PhysicsLengthUnit`, you can insert it as a resource, or simply set it while adding `PhysicsPlugins`:

```rust
fn main() {
    App::new()
        .add_plugins((
            DefaultPlugins,
            // A 2D game with 20 pixels per meter
            PhysicsPlugins::default().with_length_unit(20.0),
        ))
        .run();
}
```

### Custom Constraints

Custom constraints using XPBD are currently still possible. However, the traits and systems are now located in the `dynamics::solver::xpbd` module, and users should run `solve_constraints` in `SubstepSolverSet::SolveUserConstraints` instead of `SubstepSet::SolveUserConstraints`.

### Scheduling Changes

Several scheduling internals have been changed. For example:

- The narrow phase and `PostProcessCollisions` schedule are now run in `PhysicsStepSet::NarrowPhase` instead of `SubstepSet::NarrowPhase`.
- Integration is now run in `IntegrationSet::Velocity` and `IntegrationSet::Position` instead of `SubstepSet::Integrate`.
- `SubstepSet` has been removed.
	- The solver runs in `PhysicsStepSet::Solver`.
	- The solver's system sets are in `SolverSet`.
	- Substepping is performed in `SolverSet::Substep`.
	- The substepping loop's system sets are in `SubstepSolverSet`.

Systems running in `PostProcessCollisions` may need to be modified to account for it being moved outside of the substepping loop.


crates/avian2d/examples/collision_layers.rs
```diff
@@ -2,11 +2,15 @@
...
-use examples_common_2d::XpbdExamplePlugin;
+use examples_common_2d::ExampleCommonPlugin;
 
 fn main() {
     App::new()
-        .add_plugins((DefaultPlugins, XpbdExamplePlugin))
+        .add_plugins((
+            DefaultPlugins,
+            ExampleCommonPlugin,
+            PhysicsPlugins::default().with_length_unit(15.0),
```
janhohenheim, line 12: I think a potential new user will be a bit confused by this line.
This goes for the other instances of `with_length_unit` in the other examples as well.

crates/avian2d/examples/move_marbles.rs
```diff
@@ -2,14 +2,18 @@
...
-        .add_plugins((DefaultPlugins, XpbdExamplePlugin))
+        .add_plugins((
+            DefaultPlugins,
+            ExampleCommonPlugin,
+            PhysicsPlugins::default().with_length_unit(20.0),
+        ))
         .insert_resource(ClearColor(Color::srgb(0.05, 0.05, 0.1)))
         .insert_resource(SubstepCount(6))
-        .insert_resource(Gravity(Vector::NEG_Y * 1000.0))
+        .insert_resource(Gravity(Vector::NEG_Y * 981.0))
```
janhohenheim, line 16: A tiny bit clearer to read like this imo
```suggestion
        .insert_resource(Gravity(Vector::NEG_Y * 9.81 * 100.0))
```

src/collision/broad_phase.rs
```diff
@@ -200,7 +202,11 @@ fn sweep_and_prune(
                 continue;
             }
 
-            broad_collision_pairs.push((*ent1, *ent2));
+            if *ent1 < *ent2 {
+                broad_collision_pairs.push((*ent1, *ent2));
+            } else {
+                broad_collision_pairs.push((*ent2, *ent1));
+            }
```
janhohenheim, line 209: Too bad [`minmax`](https://doc.rust-lang.org/stable/std/cmp/fn.minmax.html) is nightly only :/

src/collision/collider/world_query.rs
```diff
@@ -0,0 +1,33 @@
+#![allow(missing_docs)]
+
+use crate::prelude::*;
+use bevy::{ecs::query::QueryData, prelude::*};
+
+/// A [`QueryData`] struct to make code handling colliders in collisions cleaner.
```
janhohenheim, line 6: Some usage examples would be nice

src/collision/collider/world_query.rs
```diff
@@ -0,0 +1,33 @@
+#![allow(missing_docs)]
```
janhohenheim, line 1: I'd scope this to the `ColliderQuery` to make sure future additions to the file are not missing docs

src/collision/collider/world_query.rs
```diff
@@ -0,0 +1,33 @@
...
+
+impl<'w, C: AnyCollider> ColliderQueryItem<'w, C> {
+    /// Returns the current position of the body. This is a sum of the [`Position`] and
+    /// [`AccumulatedTranslation`] components.
+    pub fn current_position(&self) -> Vector {
+        self.position.0
+            + self
+                .accumulated_translation
+                .as_ref()
+                .map_or(default(), |t| t.0)
```
janhohenheim, line 31:
```suggestion
                .map_or_else(default, |t| t.0)
```

src/collision/collider_backend.rs
```diff
@@ -372,13 +371,28 @@ fn update_aabb<C: AnyCollider>(
         (&Position, Option<&LinearVelocity>, Option<&AngularVelocity>),
         With<Children>,
     >,
-    dt: Res<Time>,
-    narrow_phase_config: Option<Res<NarrowPhaseConfig>>,
+    narow_phase_config: Res<NarrowPhaseConfig>,
```
janhohenheim, line 374:
```suggestion
    narrow_phase_config: Res<NarrowPhaseConfig>,
```

src/collision/collider_backend.rs
```diff
@@ -372,13 +371,28 @@ fn update_aabb<C: AnyCollider>(
...
+    length_unit: Res<PhysicsLengthUnit>,
+    time: Res<Time>,
 ) {
-    // Safety margin multiplier bigger than DELTA_TIME to account for sudden accelerations
-    let safety_margin_factor = 2.0 * dt.delta_seconds_adjusted();
+    let delta_secs = time.delta_seconds_adjusted();
+    let default_speculative_margin = length_unit.0 * narow_phase_config.default_speculative_margin;
+    let contact_tolerance = length_unit.0 * narow_phase_config.contact_tolerance;
+
+    for (collider, mut aabb, pos, rot, collider_parent, speculative_margin, lin_vel, ang_vel) in
```
janhohenheim, line 382: This seems like a good place to iterate in parallel using `for_each`

src/collision/collider_backend.rs
```diff
@@ -407,51 +421,32 @@ fn update_aabb<C: AnyCollider>(
...
+                    pos.0
+                        + (lin_vel.0 * delta_secs)
+                            .clamp_length_max(speculative_margin.max(contact_tolerance)),
                     Quaternion::from_xyzw(x, y, z, w).normalize(),
                 )
             }
         };
-
         // Compute swept AABB, the space that the body would occupy if it was integrated for one frame
+        // TODO: Should we expand the AABB in all directions for speculative contacts?
```
janhohenheim, line 448: I personally like to always link tracking issues to TODO comments, but you mileage may vary

src/collision/mod.rs
```diff
@@ -289,8 +291,15 @@ pub struct Contacts {
     /// The total tangent impulse applied to the first body in a collision.
     ///
     /// To get the corresponding force, divide the impulse by `Time<Substeps>`.
+    #[cfg(feature = "2d")]
     #[doc(alias = "total_friction_impulse")]
     pub total_tangent_impulse: Scalar,
+    /// The total tangent impulse applied to the first body in a collision.
+    ///
+    /// To get the corresponding force, divide the impulse by `Time<Substeps>`.
```
janhohenheim, line 299: I assume you mean this?
```suggestion
    /// To get the corresponding force, divide the impulse by `Time<Substeps>::delta_seconds()`.
```

src/collision/contact_query.rs
```diff
@@ -201,7 +201,8 @@ pub fn contact_manifolds(
                         normal1,
                         normal2,
                         -contact.dist,
-                        0,
+                        PackedFeatureId::UNKNOWN,
```
janhohenheim, line 204: The `ContactData` constructor is very unreadable in its current state. I'd remove the `new` call and build it explicitly. Maybe add a `Default` so that you can do `..default()` for the three field that you leave at `0.0` / `default()`?

src/collision/narrow_phase.rs
```diff
@@ -72,16 +83,16 @@ impl<C: AnyCollider> Plugin for NarrowPhasePlugin<C> {
             .get_schedule_mut(PhysicsSchedule)
             .expect("add PhysicsSchedule first");
 
+        // Manage collision states like `during_current_frame` and remove old contacts.
+        // Only one narrow phase instance should do this.
+        // TODO: It would be nice not to have collision state logic in the narrow phase.
```
janhohenheim, line 88: Add tracking issue

src/collision/narrow_phase.rs
```diff
@@ -130,26 +142,35 @@ struct NarrowPhaseInitialized;
...
 pub struct NarrowPhaseConfig {
-    /// The maximum separation distance allowed for a collision to be accepted.
+    /// The default maximum [speculative margin](SpeculativeMargin) used for
+    /// [speculative collisions](dynamics::ccd#speculative-collision). This can be overridden
+    /// for individual entities with the [`SpeculativeMargin`] component.
+    ///
+    /// This is implicitly scaled by the [`PhysicsLengthUnit`].
+    ///
+    /// Default: `MAX` (unbounded)
+    pub default_speculative_margin: Scalar,
```
janhohenheim, line 152: I feel like it's not clear for a user why they would want to override this and what it's effects are based on the current docs.

src/collision/narrow_phase.rs
```diff
@@ -130,26 +142,35 @@ struct NarrowPhaseInitialized;
...
+    /// issues and missed collisions for resting contacts.
+    ///
+    /// This is implicitly scaled by the [`PhysicsLengthUnit`].
     ///
-    /// This can be used for things like **speculative contacts** where the contacts should
-    /// include pairs of entities that *might* be in contact after constraint solving or
-    /// other positional changes.
-    pub prediction_distance: Scalar,
+    /// Default: `0.005`
+    pub contact_tolerance: Scalar,
```
janhohenheim, line 161: I feel like it's not clear for a user why they would want to override this and what it's effects are based on the current docs.
Also: what is the physical meaning of this number? Is it a random factor? Or a distance?

src/collision/narrow_phase.rs
```diff
@@ -162,139 +183,457 @@ pub enum NarrowPhaseSet {
...
+    let warm_start = solver_config.warm_start_coefficient > 0.0;
+
+    narrow_phase.update(
+        &broad_collision_pairs,
+        warm_start,
+        time.delta_seconds_adjusted(),
+    );
+}
+
+// TODO: It'd be nice to generate the constraint in the same parallel loop as `collect_collisions`
```
janhohenheim, line 207: Add tracking issue

src/collision/narrow_phase.rs
```diff
@@ -162,139 +183,457 @@ pub enum NarrowPhaseSet {
...
+        ) {
+            // At least one of the bodies must be dynamic for contact constraints
+            // to be generated.
+            if !body1.rb.is_dynamic() && !body2.rb.is_dynamic() {
+                continue;
+            }
+
+            // Generate contact constraints for the computed contacts
+            // and add them to `constraints`.
+            narrow_phase.generate_constraints(
```
janhohenheim, line 250: Can we put this into a struct so that the call becomes cleaner?

src/collision/narrow_phase.rs
```diff
@@ -162,139 +183,457 @@ pub enum NarrowPhaseSet {
...
+impl<'w, 's, C: AnyCollider> NarrowPhase<'w, 's, C> {
+    /// Updates the narrow phase by computing [`Contacts`] based on [`BroadCollisionPairs`]
+    /// and adding them to [`Collisions`].
+    fn update(
+        &mut self,
+        broad_collision_pairs: &[(Entity, Entity)],
+        warm_start: bool,
+        delta_secs: Scalar,
+    ) {
+        // TODO: These scaled versions could be in their own resource
```
janhohenheim, line 294: Add tracking issue

src/collision/narrow_phase.rs
```diff
@@ -162,139 +183,457 @@ pub enum NarrowPhaseSet {
...
-                        *entity1,
-                        *entity2,
-                        &query,
-                        &collisions,
-                        &narrow_phase_config,
-                        |contacts| {
+
+        #[cfg(feature = "parallel")]
+        {
+            // TODO: Verify if `par_splat_map` is deterministic. If not, sort the constraints (and collisions).
```
janhohenheim, line 305: Add tracking issue

src/collision/narrow_phase.rs
```diff
@@ -162,139 +183,457 @@ pub enum NarrowPhaseSet {
...
+            // Clamp velocities to the maximum speculative margins.
+            if speculative_margin1 < Scalar::MAX {
+                lin_vel1 = lin_vel1.clamp_length_max(speculative_margin1 * inv_delta_secs);
+            }
+            if speculative_margin2 < Scalar::MAX {
+                lin_vel2 = lin_vel2.clamp_length_max(speculative_margin2 * inv_delta_secs);
+            }
 
-        collisions.extend(new_collisions);
+            // TODO: Check if AABBs intersect?
```
janhohenheim, line 416: Add tracking issue

src/collision/narrow_phase.rs
```diff
@@ -162,139 +183,457 @@ pub enum NarrowPhaseSet {
...
+        &self,
+        collider1: &ColliderQueryItem<C>,
+        collider2: &ColliderQueryItem<C>,
+        max_distance: Scalar,
+        match_contacts: bool,
+    ) -> Option<Contacts> {
+        let position1 = collider1.current_position();
+        let position2 = collider2.current_position();
+
+        // TODO: It'd be good to persist the manifolds and let Parry match contacts.
```
janhohenheim, line 452: Add tracking issue

src/collision/narrow_phase.rs
```diff
@@ -162,139 +183,457 @@ pub enum NarrowPhaseSet {
...
+            .or(self
+                .collisions
+                .get_internal()
+                .get(&(collider2.entity, collider1.entity)));
+
+        let mut total_normal_impulse = 0.0;
+        let mut total_tangent_impulse = default();
+
+        // Match contacts and copy previous contact impulses for warm starting the solver.
+        // TODO: This condition is pretty arbitrary, mainly to skip dense trimeshes.
```
janhohenheim, line 478: Add tracking issue

src/collision/narrow_phase.rs
```diff
@@ -162,139 +183,457 @@ pub enum NarrowPhaseSet {
...
+            }
+        }
 
         let contacts = Contacts {
-            entity1,
-            entity2,
+            entity1: collider1.entity,
+            entity2: collider2.entity,
+            body_entity1: collider1.parent.map(|p| p.get()),
+            body_entity2: collider2.parent.map(|p| p.get()),
```
janhohenheim, line 503:
```suggestion
            body_entity1: collider1.parent.map(Parent::get),
            body_entity2: collider2.parent.map(Parent::get),
```

src/collision/narrow_phase.rs
```diff
@@ -162,139 +183,457 @@ pub enum NarrowPhaseSet {
...
+            total_normal_impulse,
+            total_tangent_impulse,
         };
 
         if !contacts.manifolds.is_empty() {
-            handle_collision(contacts);
+            return Some(contacts);
+        }
+
+        None
```
janhohenheim, line 519:
```suggestion
            Some(contacts);
        } else {
            None
        }
```

src/collision/narrow_phase.rs
```diff
@@ -162,139 +183,457 @@ pub enum NarrowPhaseSet {
...
+                Some(n) => format!("{:?} ({n})", contacts.entity1),
+                None => format!("{:?}", contacts.entity1),
+            };
+            let debug_id2 = match name2 {
+                Some(n) => format!("{:?} ({n})", contacts.entity2),
+                None => format!("{:?}", contacts.entity2),
+            };
+            warn!(
+                "{} and {} are overlapping at spawn, which can result in explosive behavior.",
+                debug_id1, debug_id2,
```
janhohenheim, line 633:
```suggestion
                "{debug_id1} and {debug_id2} are overlapping at spawn, which can result in explosive behavior."
```

src/prepare.rs
```diff
@@ -202,17 +202,8 @@ pub fn init_transforms<C: Component>(
...
-        transform,
-        global_transform,
-        pos,
-        previous_pos,
-        rot,
-        previous_rot,
-        parent,
-        has_rigid_body,
-    ) in &query
+    for (entity, transform, global_transform, pos, rot, previous_rot, parent, has_rigid_body) in
```
janhohenheim, line 205: seems like a prime candidate for a `for_each` parallel iteration

src/dynamics/solver/xpbd/mod.rs
```diff
@@ -351,3 +363,103 @@ pub fn solve_constraint<C: XpbdConstraint<ENTITY_COUNT> + Component, const ENTIT
...
+            &PreSolveRotation,
+            &mut AngularVelocity,
+        ),
+        Without<Sleeping>,
+    >,
+    time: Res<Time>,
+) {
+    let delta_secs = time.delta_seconds_adjusted();
+
+    for (rb, rot, prev_rot, mut ang_vel) in &mut bodies {
```
janhohenheim, line 447: I'd `for_each` this as well

janhohenheim (changes requested):

janhohenheim:
It would be nice to have a benchmark for this rework. It looks like things are faster, but having actual numbers would be cool.

src/collision/narrow_phase.rs
Jondolf: I used `then_some` instead
```rust
(!contacts.manifolds.is_empty()).then_some(contacts)
```

src/collision/narrow_phase.rs
Jondolf: This would be `ColliderParent::get`, not sure if it's cleaner than just the closure

src/collision/contact_query.rs
Jondolf: There aren't really default contact points or normals that would make much sense, so I'm not sure about adding a `Default` impl. One option could be to at least remove the feature IDs from the constructor and make them default to `PackedFeatureId::UNKNOWN`, and to have a `with_feature_ids` method for when you need to specify them

src/collision/collider/world_query.rs
Jondolf: This is actually on purpose due to an issue with `QueryData`. Originally `ColliderQuery` was even in a different module, but I moved it to its own module so that I can make this `allow` more scoped.

The issue is that `QueryData` also auto-generates a public `ColliderQueryItem` type, which gives the `missing_docs` warning, and I'm not aware of a way to add this `allow` for it without targeting the whole module. I should probably add a comment that mentions this though.

src/collision/collider/world_query.rs
Jondolf: This is intended more as an internal type, but I can make the docs a bit clearer and maybe add some basic examples

crates/avian2d/examples/collision_layers.rs
Jondolf: These examples have a 2D camera with a default projection, so the objects use pixels for their dimensions and are quite large in terms of physics units if the length unit was the default `1.0`. This leads to the issues with overly soft contacts and incorrectly tuned tolerances, which can be seen at the end of the "Solver Results" section (the two screenshots).

Note that bevy_rapier also has this, but in 2D it calls the method [`pixels_per_meter`](https://github.com/dimforge/bevy_rapier/blob/56f98706751cb5cf74cb82e2b73749282c8ed5ee/bevy_rapier2d/examples/boxes2.rs#L13), even though internally the property is still called `length_unit`. And it also has a `with_length_unit` method, so it just has two differently named methods that essentially configure the same thing.

A quick survey on what names different engines use for this property:

- Rapier: [`length_unit`](https://docs.rs/rapier2d/latest/rapier2d/dynamics/struct.IntegrationParameters.html#structfield.length_unit) (or in 2D, also the `pixels_per_meter` method)
- Box2D: [`b2_lengthUnitsPerMeter`](https://github.com/erincatto/box2c/blob/1d7d1cf14722a06f6a7b1c3b0850e70bcbec8558/src/core.h#L110)
- PhysX: [`PxTolerancesScale.length`](https://nvidia-omniverse.github.io/PhysX/physx/5.1.0/docs/API.html#using-different-units)

I agree though that the method usage should probably be commented in the examples, or the method name itself could be more self-explanatory. It's just hard to come up with an approachable name that conveys the purpose, but does *not* imply that it actually scales velocities or forces in any way.

For example, if there was a `physics_scale` property and it was set to `10.0` (let's imagine this is 2D), a user could reasonably assume that `Collider::rectangle(1.0, 1.0)` would get scaled to a 10x10 pixel collider. Or you could assume that you could use `LinearVelocity::X`, and have the body move at the speed of 10 pixels per second. But this is not what `length_unit` does, as it just scales the internal length-based tolerances and thresholds to match the expected scale. The nuance is difficult to convey through the name of the property.

For these examples though, another option would be to just scale the camera projection so that physics objects don't have to be so large in terms of physics units and you don't need a custom length unit. This would technically be more ideal for physics in other ways too, although I don't know what the game dev best practices are for 2D games when it comes to using pixels vs. other units for scenes.

(sorry for the long response lol)

src/collision/collider_backend.rs
Jondolf: I got mixed results when running the benchmarks, same for the other `for_each` suggestions. Overall it seems like it just causes performance regressions due to the added overhead, especially for smaller scenes.

With a scene that has a lot more physics objects but less active collisions, it might be more worth it, but I think we would need to trace a more realistic large-scale game scene to be certain, and we don't really have that currently. So I think I'll leave it for now, but it might be worth looking into and benchmarking further in the future.

src/collision/narrow_phase.rs
janhohenheim: Fair enough

src/collision/contact_query.rs
Jondolf: For now I just removed the feature IDs from the constructor and added a `with_feature_ids` method. I don't think the constructor is *too* unreadable, it's just two points, two normals, and a penetration depth. It would be even clearer if we used `Dir2`/`Dir3` for the normals though (I might change it to this in the future)

src/collision/narrow_phase.rs
Jondolf: I tried to make this a bit clearer in [c86797f](https://github.com/Jondolf/bevy_xpbd/pull/385/commits/c86797f334de1dec1a7384068bbba8fd75097161)

src/collision/narrow_phase.rs
Jondolf: I tried to make this a bit clearer in [c86797f](https://github.com/Jondolf/bevy_xpbd/pull/385/commits/c86797f334de1dec1a7384068bbba8fd75097161)

src/collision/narrow_phase.rs
Jondolf: Not sure if it's worth it unless it could be reused elsewhere. I'm also not sure what the struct could be called

Jondolf:
Addressed most of the comments. A few other changes:

- `IntegratorPlugin` initializes its system sets in the schedule given to it instead of *only* configuring them in `SolverPlugin`. This way, it would work on its own even without the solver.
- Added a `match_contacts` property to `NarrowPhaseConfig`. Previously it was decided by `SolverConfig::warm_start_coefficient`, but that's a solver implementation detail.
- `NarrowPhasePlugin` has a `generate_constraints` property. It's unnecessary to generate constraints if you only use the narrow phase for collision detection without the solver.
- `NarrowPhasePlugin` can be initialized with any schedule, not just `PhysicsSchedule`.

crates/avian2d/examples/collision_layers.rs
Jondolf: For now, I added a comment briefly explaining the length unit to each example that sets it. I also improved the length unit docs overall

janhohenheim (approved):

PR #386, opened 2024-06-25
Support default layers and density for `ColliderConstructorHierarchy`

Jondolf:
# Objective

When loading a (glTF) scene and generating colliders for it, it can be useful to be able to add all entities in that scene to some collision layer. However, `ColliderConstructorHierarchy` (previously `AsyncSceneCollider`, see #378) doesn't currently allow specifying default collision layer or density configurations despite allowing configurations for individual entities.

## Solution

Add `default_layers` and `default_density` properties for `ColliderConstructorHierarchy`. The `layers` and `density` properties of `ColliderConstructorHierarchyData` are now `Option`s, and if they are `None`, the default values are used.

I also moved the collider constructor types and tests to their own module, and updated some docs to remove mesh-specific wording.

For reviewing purposes, the first commit (8cbe48c) should be easier to read, the other commit just moves code to the `constructor` module.


src/plugins/collision/collider/constructor.rs
```diff
@@ -0,0 +1,788 @@
...
+    }
+}
+
+/// Configuration for a specific collider generated from a scene using [`ColliderConstructorHierarchy`].
+#[derive(Clone, Debug, PartialEq, Reflect)]
+#[reflect(Debug, PartialEq)]
+#[cfg_attr(feature = "serialize", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serialize", reflect(Serialize, Deserialize))]
+#[cfg_attr(all(feature = "3d", feature = "collider-from-mesh"), reflect(Default))]
+pub struct ColliderConstructorHierarchyConfig {
```
janhohenheim, line 234: It's weird how everything but the constructor is optional. With this PR, wouldn't it make sense to want to override only the `layers` and leave the default constructor? I'd suggest making the constructor an `Option` as well.

This will also remove the special-cased `Default` and the need for `ColliderConstructorHierarchyConfig::from_constructor`

janhohenheim (changes requested):

src/plugins/collision/collider/constructor.rs
Jondolf: Yeah, I agree. I made the constructor an `Option` and reworked the related logic a bit

janhohenheim (changes requested):
Feature looks good to me now :) Could you duplicate the `collider_constructor_hierarchy_inserts_computed_colliders_on_scene` test and adapt it to test the new behavior?

janhohenheim (approved):

PR #393, opened 2024-07-02
Implement collision margin

Jondolf:
# Objective

Collisions for thin objects such as triangle meshes can sometimes have stability and performance issues, especially when dynamically colliding against other thin objects.

- Infinitely thin objects are more prone to numerical issues and tunneling. Two trimeshes colliding against each other can easily get stuck in an intersecting state or have other collision issues.
- Collision algorithms for intersecting objects are typically more expensive than for the non-intersecting state. Trimesh collisions can be very expensive if the trimesh has dense geometry.

To allow users to mitigate these issues when necessary, engines such as Rapier and Bullet provide a *collision margin* or *contact skin* that forces the solver to maintain an artificial separation distance between objects. It can be thought of as a kind of shell that adds thickness to colliders, reducing the issues mentioned above.

## Solution

Add a `CollisionMargin` component. It can be used to add artifical thickness to any collider for collisions.

```rust
commands.spawn((
    RigidBody::Dynamic,
    Collider::trimesh_from_mesh(&mesh),
    CollisionMargin(0.05),
));
```

Without a collision margin:

https://github.com/Jondolf/bevy_xpbd/assets/57632562/0364168b-81e9-4887-a2fa-10a23b87481e

With a collision margin:

https://github.com/Jondolf/bevy_xpbd/assets/57632562/b41a3cbc-0709-4c3f-9f20-1b73d2d48e76

### DIFF
--- a/src/collision/collider/mod.rs
+++ b/src/collision/collider/mod.rs
@@ -341,9 +341,9 @@ impl Default for ColliderAabb {
     }
 }
 
-/// A component that adds an extra margin around [`Collider`] shapes to help maintain
-/// additional separation to other objects. This can be imagined as an extra shell
-/// or skin that adds artificial thickness.
+/// A component that adds an extra margin or "skin" around [`Collider`] shapes to help maintain
+/// additional separation to other objects. This added thickness can help improve
+/// stability and performance in some cases, especially for thin shapes such as trimeshes.
 ///
 /// There are three primary reasons for collision margins:
 ///
@@ -395,8 +395,9 @@ impl Default for ColliderAabb {
 )]
 /// }
 /// ```
-#[derive(Reflect, Clone, Component, Debug, Default, Deref, DerefMut, PartialEq)]
+#[derive(Reflect, Clone, Copy, Component, Debug, Default, Deref, DerefMut, PartialEq)]
 #[cfg_attr(feature = "serialize", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serialize", reflect(Serialize, Deserialize))]
 #[reflect(Component)]
 #[doc(alias = "ContactSkin")]
 pub struct CollisionMargin(pub Scalar);
--- a/src/collision/collider/mod.rs
+++ b/src/collision/collider/mod.rs
@@ -389,7 +389,7 @@ impl Default for ColliderAabb {
     // A margin of `0.1` is added around the shape.
     commands.spawn((
         RigidBody::Dynamic,
-        Collider::trimesh_from_mesh(&mesh),
+        Collider::trimesh_from_mesh(&mesh).unwrap(),
         CollisionMargin(0.1),
     ));"
 )]
--- a/src/collision/collider/mod.rs
+++ b/src/collision/collider/mod.rs
@@ -6,6 +6,7 @@ use bevy::{
     prelude::*,
     utils::HashSet,
 };
+use derive_more::From;
 
 mod backend;
 mod hierarchy;
@@ -395,7 +396,7 @@ impl Default for ColliderAabb {
 )]
 /// }
 /// ```
-#[derive(Reflect, Clone, Copy, Component, Debug, Default, Deref, DerefMut, PartialEq)]
+#[derive(Reflect, Clone, Copy, Component, Debug, Default, Deref, DerefMut, PartialEq, From)]
 #[cfg_attr(feature = "serialize", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "serialize", reflect(Serialize, Deserialize))]
 #[reflect(Component)]
--- a/src/collision/collider/mod.rs
+++ b/src/collision/collider/mod.rs
@@ -341,6 +341,66 @@ impl Default for ColliderAabb {
     }
 }
 
+/// A component that adds an extra margin around [`Collider`] shapes to help maintain
+/// additional separation to other objects. This can be imagined as an extra shell
+/// or skin that adds artificial thickness.
+///
+/// There are three primary reasons for collision margins:
+///
+/// 1. Collision detection is often more efficient when shapes are not overlapping
+/// further than their collision margins. Deeply overlapping shapes require
+/// more expensive collision algorithms.
+///
+/// 2. Some shapes such as triangles and planes are infinitely thin,
+/// which can cause precision errors. A collision margin adds artificial
+/// thickness to shapes, improving stability.
+///
+/// 3. Overall, collision margins give the physics engine more
+/// room for error when resolving contacts. This can also help
+/// prevent visible artifacts such as objects poking through the ground.
+///
+/// If a rigid body with a [`CollisionMargin`] has colliders as child entities,
+/// and those colliders don't have their own [`CollisionMargin`] components,
+/// the colliders will use the rigid body's [`CollisionMargin`].
+///
+/// # Example
+///
+/// ```
+#[cfg_attr(feature = "2d", doc = "use avian2d::prelude::*;")]
+#[cfg_attr(feature = "3d", doc = "use avian3d::prelude::*;")]
+/// use bevy::prelude::*;
+///
+/// fn setup(mut commands: Commands) {
+#[cfg_attr(
+    feature = "2d",
+    doc = "    // Spawn a rigid body with a collider.
+    // A margin of `0.1` is added around the shape.
+    commands.spawn((
+        RigidBody::Dynamic,
+        Collider::capsule(2.0, 0.5),
+        CollisionMargin(0.1),
+    ));"
+)]
+#[cfg_attr(
+    feature = "3d",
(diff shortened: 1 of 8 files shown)


src/collision/collider/mod.rs
```diff
@@ -341,6 +341,66 @@ impl Default for ColliderAabb {
...
+)]
+#[cfg_attr(
+    feature = "3d",
+    doc = "    let mesh = Mesh::from(Torus::default());
+
+    // Spawn a rigid body with a triangle mesh collider.
+    // A margin of `0.1` is added around the shape.
+    commands.spawn((
+        RigidBody::Dynamic,
+        Collider::trimesh_from_mesh(&mesh).unwrap(),
```
janhohenheim, line 392: Any particular reason you're not using a primitive here? It looks to me like using a trimesh is making the example less readable.

src/collision/collider/mod.rs
```diff
@@ -341,6 +341,66 @@ impl Default for ColliderAabb {
...
+        CollisionMargin(0.1),
+    ));"
+)]
+/// }
+/// ```
+#[derive(Reflect, Clone, Component, Debug, Default, Deref, DerefMut, PartialEq)]
+#[cfg_attr(feature = "serialize", derive(serde::Serialize, serde::Deserialize))]
```
janhohenheim, line 399:
```suggestion
#[cfg_attr(feature = "serialize", derive(serde::Serialize, serde::Deserialize))]
#[cfg_attr(feature = "serialize", reflect(Serialize, Deserialize))]
```

src/collision/collider/mod.rs
```diff
@@ -341,6 +341,66 @@ impl Default for ColliderAabb {
...
+    // A margin of `0.1` is added around the shape.
+    commands.spawn((
+        RigidBody::Dynamic,
+        Collider::trimesh_from_mesh(&mesh).unwrap(),
+        CollisionMargin(0.1),
+    ));"
+)]
+/// }
+/// ```
+#[derive(Reflect, Clone, Component, Debug, Default, Deref, DerefMut, PartialEq)]
```
janhohenheim, line 398:
```suggestion
#[derive(Reflect, Clone, Copy, Component, Debug, Default, Deref, DerefMut, PartialEq)]
```

src/collision/narrow_phase.rs
```diff
@@ -279,16 +279,31 @@ fn generate_constraints<C: AnyCollider>(
...
+            // The collision margin adds artificial thickness to colliders for performance
+            // and stability. See the `CollisionMargin` documentation for more details.
+            let collision_margin1 = collider1.collision_margin.map_or(
+                rb_collision_margin1.map_or(0.0, |margin| margin.0),
+                |margin| margin.0,
+            );
+            let collision_margin2 = collider2.collision_margin.map_or(
+                rb_collision_margin2.map_or(0.0, |margin| margin.0),
+                |margin| margin.0,
+            );
```
janhohenheim, line 304:
```suggestion
            let collision_margin1 = collider1.collision_margin
                .or(rb_collision_margin1)
                .unwrap_or_default().0;
            let collision_margin2 = collider2.collision_margin
                .or(rb_collision_margin2)
                .unwrap_or_default().0;
```

The `map_or` chains are very unreadable imo, so I'd rewrite them like this

src/collision/narrow_phase.rs
```diff
@@ -552,6 +608,7 @@ impl<'w, 's, C: AnyCollider> NarrowPhase<'w, 's, C> {
         body2: &RigidBodyQueryReadOnlyItem,
         collider1: &ColliderQueryItem<C>,
         collider2: &ColliderQueryItem<C>,
+        collision_margin: Scalar,
```
janhohenheim, line 611:
```suggestion
        collision_margin: CollisionMargin,
```
It's weird to refer to `CollisionMargin` in the docs and then not actually use that type.

src/collision/narrow_phase.rs
```diff
@@ -399,16 +423,43 @@ impl<'w, 's, C: AnyCollider> NarrowPhase<'w, 's, C> {
...
+            body2_bundle.as_ref().map_or(
+                (Vector::ZERO, None, None),
+                |(body, collision_margin, speculative_margin)| {
+                    (
+                        body.linear_velocity.0,
+                        *collision_margin,
+                        *speculative_margin,
+                    )
+                },
+            );
```
janhohenheim, line 447:
```suggestion
        let (mut lin_vel1, rb_collision_margin1, rb_speculative_margin1) = body1_bundle
            .as_ref()
            .map(|(body, collision_margin, speculative_margin)| {
                (
                    body.linear_velocity.0,
                    *collision_margin,
                    *speculative_margin,
                )
            })
            .unwrap_or_default();
        let (mut lin_vel2, rb_collision_margin2, rb_speculative_margin2) = body2_bundle
            .as_ref()
            .map(|(body, collision_margin, speculative_margin)| {
                (
                    body.linear_velocity.0,
                    *collision_margin,
                    *speculative_margin,
                )
            })
            .unwrap_or_default();
```

src/dynamics/solver/contact/mod.rs
```diff
@@ -91,6 +91,7 @@ impl ContactConstraint {
         collider_entity2: Entity,
         collider_transform1: Option<ColliderTransform>,
         collider_transform2: Option<ColliderTransform>,
+        collision_margin: Scalar,
```
janhohenheim, line 94:
```suggestion
        collision_margin: CollisionMargin,
```
if we already have the newtypes, we should also be consistently using them imo

src/dynamics/ccd/mod.rs
```diff
@@ -216,6 +216,11 @@
 //! For shapes that are intended to be solid from the inside, it is recommended
 //! to use convex decomposition instead.
 //!
+//! If you must use triangle mesh colliders and are having stability issues, consider
```
janhohenheim, line 219: Is this only relevant for CCD?

src/collision/narrow_phase.rs
```diff
@@ -399,16 +423,43 @@ impl<'w, 's, C: AnyCollider> NarrowPhase<'w, 's, C> {
...
+        // and stability. See the `CollisionMargin` documentation for more details.
+        let collision_margin1 = collider1.collision_margin.map_or(
+            rb_collision_margin1.map_or(0.0, |margin| margin.0),
+            |margin| margin.0,
+        );
+        let collision_margin2 = collider2.collision_margin.map_or(
+            rb_collision_margin2.map_or(0.0, |margin| margin.0),
+            |margin| margin.0,
+        );
+        let collision_margin_sum = collision_margin1 + collision_margin2;
```
janhohenheim, line 462: This code is duplicated. Consider extracting it into a method.

janhohenheim (changes requested):

janhohenheim:
Math checks out, just some code quality issues, nothing big :)

src/collision/collider/mod.rs
Jondolf: Collision margins are most useful for trimeshes. Primitives are typically not infinitely thin or hollow, and collisions against them are much cheaper and generally stabler, so using them here wouldn't perhaps be as representative of common usage.

src/dynamics/ccd/mod.rs
Jondolf: This section in the CCD docs is under an "Other Ways to Avoid Tunneling" heading. But I'll mention more explicitly in the `CollisionMargin` docs that it can be especially useful for trimesh colliders

janhohenheim: I'd add it to the trimesh constructor itself, since otherwise a user wouldn't stumble upon the fix without knowing about it already.

src/collision/narrow_phase.rs
Jondolf: You can't `unwrap_or_default` on `&CollisionMargin` (a reference), so it would become this:
```rust
let collision_margin1 = collider1
    .collision_margin
    .or(rb_collision_margin1)
    .copied()
    .unwrap_or_default()
    .0;
let collision_margin2 = collider2
    .collision_margin
    .or(rb_collision_margin2)
    .copied()
    .unwrap_or_default()
    .0;
```
I don't think that's better

janhohenheim: You could swap the `.copied().unwrap_or_default()` for a `.map_or(0.0, |margin| margin.0)`, but I very much would appreciate leaving the `or` for readability. Maybe it's just me, but chained `map_or`s induce way more mental load for me than function chains.

src/collision/narrow_phase.rs
Jondolf: I don't like having to wrap everything in newtypes unnecessarily so I changed this (and the other similar cases) to `impl Into<CollisionMargin>` and derived `From<Scalar>` for now. Not sure if it's worth it though, since the newtypes don't really give an API benefit (other than explicitness and docs) and it just means that we need to call `.into()` and use derefs in more places internally

PR #402, opened 2024-07-06
Take `SpatialQueryFilter` by reference in spatial queries

Jondolf:
# Objective

Spatial queries currently take `SpatialQueryFilter` by value and clone it several times. This could sometimes be cloning a pretty hefty amount of data if the `HashSet` of excluded entities is large.

## Solution

Take spatial query filters by reference and reduce the amount of cloning.

## Future Work

Shapecasting with multiple hits should be optimized to only traverse once instead of doing best-first traversals in a loop. See #403.

---

## Migration Guide

Spatial queries performed through `SpatialQuery` now take `SpatialQueryFilter` by reference.


janhohenheim (approved):
Oh nice, you managed to make all of the spatial queries take the filter by reference!

PR #405, opened 2024-07-07
Fix `Mesh` not being optional

Jondolf:
# Objective

Avian 0.1.0 introduced a regression where `bevy_render` is now needed because `Mesh` is not optional. It should be feature-gated behind `collider-from-mesh`.

## Solution

Put all mentions of `Mesh` behind the `collider-from-mesh` feature.


janhohenheim (changes requested):
[`collider_from_mesh` requires `3d`](https://github.com/Jondolf/avian/blob/main/crates/avian3d/Cargo.toml#L44), so you can change all instances of `#[cfg(all(feature = "3d", feature = "collider-from-mesh"))]` to `#[cfg(feature = "collider-from-mesh")]`

janhohenheim (approved):

PR #413, opened 2024-07-10
Fix remove ancestor markers crash

JonasAAA:
# Objective

Fix the issue where `remove_ancestor_markers` crashes if an entity has children moved or removed, and then is despawned before `remove_ancestor_markers`  is run.

## Solution

When remove component from an entity, just check if that entity exists first.


Jondolf (approved):
Thanks, good catch! I appreciate the regression test as well :)

PR #421, opened 2024-07-13
Fixed add_angle

hocop:
# Objective

Fixes [415](https://github.com/Jondolf/avian/issues/415) and [416](https://github.com/Jondolf/avian/issues/416)

## Solution# Objective

Old implementation was only for small angles. When impact on `RevoluteJoint` is big, XPBD [tries to apply](https://github.com/Jondolf/avian/blob/main/src/dynamics/solver/xpbd/positional_constraint.rs#L70) a big update on angles and fails

## Solution

Implemented `add_angle` function using precise computation of sin and cos

### DIFF
--- a/src/position.rs
+++ b/src/position.rs
@@ -408,8 +408,7 @@ impl Rotation {
     #[must_use]
     /// Adds the given counterclockiwise angle in radians to the [`Rotation`].
     pub fn add_angle(&self, radians: Scalar) -> Self {
-        Rotation::from_sin_cos(self.sin + radians * self.cos, self.cos - radians * self.sin)
-            .normalize()
+        Rotation::radians(radians) * *self
     }
 
     /// Performs a linear interpolation between `self` and `rhs` based on


hocop:
Although I get that the linearized implementation of `add_angle` makes more sense for XBPD updates, as they are linear.

Maybe then add another function `add_angle_linearized` and remove assert from it?

Khanoto:
I don't understand, why are you squaring the radian angle by self ?

`Rotation::radians(radians) * *self`

Khanoto:
Shouldn't we use this formula to add angle ?
image

I think something like that may do it
image

Jondolf:
The existing implementation was based on Solver2D's [`s2IntegrateRot`](https://github.com/erincatto/solver2d/blob/450fbd338c83e19e66f978c6992629a7c4a285a4/include/solver2d/math.h#L209-L229).

We should not compute it by computing `sin` and `cos` like that, as it is more expensive, and `sin` and `cos` also produce varying results on different architectures, so this would need to use `libm` for cross-platform determinism when `enhanced-determinism` is enabled, which is also more expensive.

We should just do the normalization in an optimized way before using `from_sin_cos`, similarly to what [Box2D V3](https://github.com/erincatto/box2c/blob/1d7d1cf14722a06f6a7b1c3b0850e70bcbec8558/include/box2d/math_functions.h#L306) does.

```rust
/// Adds the given counterclockiwise angle in radians to the [`Rotation`].
#[inline]
#[must_use]
pub fn add_angle(&self, radians: Scalar) -> Self {
    let (sin, cos) = (self.sin + radians * self.cos, self.cos - radians * self.sin);
    let magnitude_squared = sin * sin + cos * cos;
    let magnitude_recip = if magnitude_squared > 0.0 {
        magnitude_squared.sqrt().recip_or_zero()
    } else {
        0.0
    };
    Rotation::from_sin_cos(sin * magnitude_recip, cos * magnitude_recip)
}
```

hocop:
Thanks! I changed to your implementation and checked that it works

hocop:
@Khanoto I think it is identical to what I initially proposed, just more verbose

Jondolf (approved):
Thanks!

PR #468, opened 2024-07-23
Disable overlap warning for sensors

hocop:
# Objective

There is a warning that looks like this:
```
WARN avian2d::collision::narrow_phase: Entity { index: 8, generation: 1 } and Entity { index: 20, generation: 1 } are overlapping at spawn, which can result in explosive behavior.
```

In my opinion it shouldn't show up when we are spawning `Sensor`s. They may overlap without explosive behavior.

## Solution

Sensors are filtered out from the `log_overlap_at_spawn` system

### DIFF
--- a/src/collision/narrow_phase.rs
+++ b/src/collision/narrow_phase.rs
@@ -689,7 +689,7 @@ impl<'w, 's, C: AnyCollider> NarrowPhase<'w, 's, C> {
 #[cfg(debug_assertions)]
 fn log_overlap_at_spawn(
     collisions: Res<Collisions>,
-    added_bodies: Query<(Ref<RigidBody>, Option<&Name>, &Position)>,
+    added_bodies: Query<(Ref<RigidBody>, Option<&Name>, &Position), Without<Sensor>>,
 ) {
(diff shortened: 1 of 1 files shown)


src/collision/narrow_phase.rs
```diff
@@ -689,7 +689,7 @@ impl<'w, 's, C: AnyCollider> NarrowPhase<'w, 's, C> {
 #[cfg(debug_assertions)]
 fn log_overlap_at_spawn(
     collisions: Res<Collisions>,
-    added_bodies: Query<(Ref<RigidBody>, Option<&Name>, &Position)>,
+    added_bodies: Query<(Ref<RigidBody>, Option<&Name>, &Position), Without<Sensor>>,
```
janhohenheim, line 692: Hmm? Wouldn't the `Sensor` usually be part of a `Collider`, which may or may not coincide with being a `RigidBody`?
I guess this change helps with the case where the `RigidBody` is holding both the `Collider` and the `Sensor`, but does not help with a `RigidBody` that is the ancestor of a `Collider` with a `Sensor`.

PR #470, opened 2024-07-25
Allow not overwriting entity visibility with debug renderer

Dentosal:
# Objective

Currently `PhysicsDebugPlugin` always overwrites `Visibility` component of the entity with `Collider`. This is sometimes undesirable, as I want to have entities whose visibilty is part of the game logic. 

## Solution

This PR allows the debug renderer to instead not overwrite the visibilty when configured to do so. The PR also changes the default value to not overwrite, as it always forced colliders to be visibile before.

## Alternatives

An `Option<Visibility>` could be used instead of an enum, but I hope this version is more readable. We could also just not change visibility if `hide_meshes` is set to `false`, but this would unexpectedly affect visibility if this is toggled.

---

## Changelog

- What changed as a result of this PR?
- If applicable, organize changes under "Added", "Changed", or "Fixed" sub-headings
- Stick to one or two sentences. If more detail is needed for a particular change, consider adding it to the "Solution" section
  - If you can't summarize the work, your change may be unreasonably large / unrelated. Consider splitting your PR to make it easier to review and merge!

## Migration Guide

The `hide_meshes: bool` field of `PhysicsGizmos` (global) and `DebugRender` (entity) has been replaced with `mesh_visibility: MeshVisibility`. The old value `false` should be `MeshVisibility::Overwrite(Visibility::Visible)`, and `true` should be `MeshVisibility::Overwrite(Visibility::Hidden)`.

The debug renderer no longer forces entities with `Collider` to be visible by default To restore old behavior, we have to set the `mesh_visibility` field to one of the variants described above.

### DIFF
--- a/src/debug_render/configuration.rs
+++ b/src/debug_render/configuration.rs
@@ -83,9 +83,8 @@ pub struct PhysicsGizmos {
     pub shapecast_point_color: Option<Color>,
     /// The color used for the hit normals in [shapecasts](spatial_query#shapecasting).
     pub shapecast_normal_color: Option<Color>,
-    /// Determines if the visibility of entities with [colliders](Collider) should be set to `Visibility::Hidden`,
-    /// which will only show the debug renders.
-    pub hide_meshes: bool,
+    /// Determines if the visibility of entities with [colliders](Collider).
+    pub mesh_visibility: MeshVisibility,
 }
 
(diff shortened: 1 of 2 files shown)


Dentosal:
> Marking as Waiting-on-Author due to merge conflicts

Reworked the conflicting segments, should be good to go again

PR #480, opened 2024-08-04
Make contacts deterministic across Worlds

cBournhonesque:
# Objective

- Fixes https://github.com/Jondolf/avian/issues/406

We noticed an issue where the Collisions were using the order of `Entity` directly to sort the pair of entities in a contact.
The issue is that for networking, there is no guarantee that the pair of Entities that are replicated between a Client and a Server world are in the same direction.

## Solution

- A solution is to keep using the `aabb.min.x` to order the two entities.
- There are a couple concerns:
  - we want to make sure that when we match with previous entity pairs, this is not an issue.
    - it's not a problem for `previous_contacts` here: https://github.com/cBournhonesque/avian/blob/ac1fd0269e04f19f6892c62019cbb9ad9c19444e/src/collision/narrow_phase.rs#L549 because we check both directions
    - it is a problem for `match_manifolds` here: https://github.com/cBournhonesque/avian/blob/ac1fd0269e04f19f6892c62019cbb9ad9c19444e/src/collision/narrow_phase.rs#L571 which is why we modify the function
  - we want to make sure that this doesn't change how `CollisionStarted` and `CollisionEnded` events are sent, even if the entities change their order during a contact. **we should probably add a test for this**
    - it should be fined because those events are based on the `Contacts` struct that is generated during the narrow phase, and we made sure that contact matching still works independently of past entity order

## Test

- tested in lightyear that the simulations are now completely deterministic
- probably need to test that contact matching works correctly, even if the `aabb.min.x` of the two entities gets swapped during the contact


cBournhonesque:
Maybe the docs should be improved as well to make clear that:
- Collisions has the keys in **entity order**
- Contacts has the entities in **aabb** order (and it should probably use aabb.min.y as secondary sorting key?)

Jondolf (approved):

PR #493, opened 2024-08-21
Add predicate variants to all casts

janhohenheim:
# Objective

-  `cast_ray` already has a sister function called `cast_ray_predicate` that returns the first hit that satisfies a predicate. The other casts do not.

## Solution

- Implement them. Note that I set the internal implementations of the non-predicate versions to use the predicate to reduce code duplication. The predicate does not introduce an extra branch since we already have an `if` for the filter in place, so this should be equivalent in performance.

---

## Changelog

- Added
  - `cast_ray` already has a sister function called `cast_ray_predicate` that returns the first hit that satisfies a predicate. Now, the other cast function have a similar analogue:
	  - Added `cast_ray_predicate` for `cast_ray`
	  - Added `project_point_predicate` for `project_point`

### DIFF
--- a/src/spatial_query/pipeline.rs
+++ b/src/spatial_query/pipeline.rs
@@ -169,22 +169,14 @@ impl SpatialQueryPipeline {
         solid: bool,
         query_filter: &SpatialQueryFilter,
     ) -> Option<RayHitData> {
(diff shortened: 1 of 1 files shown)


Jondolf (approved):
Thanks!

PR #494, opened 2024-08-22
Make `PhysicsLayer` require a `#[default]` variant

Jondolf:
# Objective

#476 changed collision layers to reserve the first bit for a default layer. However, for enums implementing `PhysicsLayer`, the default layer is rather implicit and potentially footgunny.

The default layer should ideally be more explicit.

## Solution

Change `PhysicsLayer` to require `Default` to be implemented. The proc macro orders the variants such that the layer set as the default is always the first bit.

Note that `Default` *must* currently be derived instead of implemented manually, as the macro can only access the default variant if the `#[default]` attribute exists. An error is emitted if `#[default]` does not exist. If someone knows a way to make it work with the manual impl, let me know or consider opening a PR!

I also added documentation for the `PhysicsLayer` macro, and improved error handling.

---

## Migration Guide

Enums that derive `PhysicsLayer` must now also derive `Default` and specify a default layer using the `#[default]` attribute.

Before:

```rust
#[derive(PhysicsLayer)]
enum GameLayer {
    Player,
    Enemy,
    Ground,
}
```

After:

```rust
#[derive(PhysicsLayer, Default)]
enum GameLayer {
    #[default]
    Default, // The name doesn't matter, but Default is used here for clarity
    Player,
    Enemy,
    Ground,
}
```

The default layer always has the bit value `0b0001` regardless of its order relative to the other variants.

### DIFF
--- a/crates/avian_derive/src/lib.rs
+++ b/crates/avian_derive/src/lib.rs
@@ -5,16 +5,48 @@ use proc_macro::TokenStream;
 use quote::{quote, quote_spanned};
 use syn::{parse_macro_input, spanned::Spanned, Data, DeriveInput};
 
-// Modified macro From the discontinued Heron,
-// see https://github.com/jcornaz/heron/blob/main/macros/src/lib.rs
+// Modified macro from the discontinued Heron
+// https://github.com/jcornaz/heron/blob/main/macros/src/lib.rs
+/// A derive macro for defining physics layers using an enum.
+///
+/// Each variant of the enum represents a layer. Each layer has a unique bit determined by
+/// the order of the variants. The bit value can be retrieved using the `to_bits` method.
+///
+/// # Requirements
+///
+/// - The enum must have at most 32 variants.
+/// - The enum variants must not have any fields.
+/// - The enum must have a default variant with the `#[default]` attribute.
+///   - The first bit `1 << 0` will *always* be reserved for the default layer.
+///     The bit values of the other layers are determined by their order in the enum, starting from `1 << 1`.
+///
+/// # Example
+///
+/// ```ignore
+/// #[derive(PhysicsLayer, Default)]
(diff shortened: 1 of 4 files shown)


crates/avian_derive/src/lib.rs
```diff
@@ -2,19 +2,51 @@
...
+///
+/// - The enum must have at most 32 variants.
+/// - The enum variants must not have any fields.
+/// - The enum must have a default variant with the `#[default]` attribute.
+///   - The first bit `1 << 0` will *always* be reserved for the default layer.
+///     The bit values of the other layers are determined by their order in the enum, starting from `1 << 1`.
+///
+/// # Example
+///
+/// ```ignore
```
janhohenheim, line 26: Why is this ignored?

Jondolf: This is in `avian_derive`, which doesn't have the `PhysicsLayer` trait from `avian2d`/`avian3d` (I think it'd cause a circular dependency), so the test would fail.

janhohenheim: Ah heck! Could you add the example as a unit test to avian then?

Jondolf: Yup! Just added an equivalent test for `LayerMask` docs for now

crates/avian_derive/src/lib.rs
```diff
@@ -2,19 +2,51 @@
...
+/// #[derive(PhysicsLayer, Default)]
```
janhohenheim, line 27:
```suggestion
/// #[derive(PhysicsLayer, Clone, Copy, Debug, Default)]
```
to keep parity with the test you just added.

Aceeri (approved):

janhohenheim (approved):

PR #497, opened 2024-08-23
revert schedule change for debug render. fixes #496

### LINKED ISSUE #496: really bad debug-render performance
major performance issues when debug rendering is enabled after commit 4d082a79

a quick look and i noticed that the gizmo systems were changed too run in FixedPostUpdate
changing this back too PostUpdate solved the lag when rendering is enabled

im not really sure if this is intended behavior of gizmos though, it does not seem correct from my understanding

Hellzbellz123:
*opened as draft cause im not sure if a whole new pr/issue is the way to go*

revert schedule change for PhysicsDebugPlugin because it causes large lag

# Objective
fixes issue #496
recent scheduling changes from commit 4d082a79 causes PhysicsDebugPlugin too generate lots of lag when enabled

## Solution
change default schedule from FixedPostUpdate too PostUpdate

---


Jondolf (approved):
Tested this, and performance for the debug rendering / gizmos indeed seems *much* worse in fixed timestep schedules 🤔 Having it in `PostUpdate` should be fine though, thanks for the fix!

For Bevy 0.16, we will likely also start using the retained gizmos added in bevyengine/bevy#15473 for collider debug rendering. It should hopefully improve performance significantly.

PR #511, opened 2024-09-08
Joint entities rotations

exoexo-dev:
# Objective

- Allow to set the local rotation of the connected entities in `RevoluteJoint` (in 3D only) and in `SphericalJoint`

## Solution

- The rotation of the entities is controlled by the joints, which forces you to nest the connected entities inside another transform in order to change their rotation relative to the joints axes
- This PR adds `local_rotation1` and `local_rotation2` (analogous to `local_anchor`s) that are being added to the rotations synced from the entities transforms
- These fields allow to set additional rotations of the entities relative to the joints axes

### DIFF
--- a/src/position.rs
+++ b/src/position.rs
@@ -754,6 +754,22 @@ impl Rotation {
 }
 
 #[cfg(feature = "3d")]
+impl std::ops::Mul for Rotation {
+    type Output = Self;
+
+    fn mul(self, rhs: Self) -> Self::Output {
+        Self(self.0 * rhs.0)
+    }
+}
+
+#[cfg(feature = "3d")]
+impl std::ops::MulAssign for Rotation {
+    fn mul_assign(&mut self, rhs: Self) {
+        *self = *self * rhs;
+    }
+}
+
+#[cfg(feature = "3d")]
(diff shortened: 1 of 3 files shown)


src/position.rs
```diff
@@ -712,6 +712,22 @@ impl Rotation {
...
+impl std::ops::Mul for Rotation {
+    type Output = Self;
+
+    fn mul(self, rhs: Self) -> Self::Output {
+        Self(self.0 * rhs.0)
+    }
+}
+
+#[cfg(feature = "3d")]
+impl std::ops::MulAssign for Rotation {
```
exoexo-dev, line 725: this impl is not needed for the impl, but it probably doesn't hurt to have it

Jondolf:
Hi, sorry for leaving this sitting for so long :sweat_smile: I wanted to implement this more consistently for *all* joints using local frames (anchor + basis), which are commonly used for this sort of thing in most other engines such as Rapier, Box2D (main branch), Bullet, PhysX...

I've implemented that as part of #803 for all joints (except the `DistanceJoint` which only has anchors), so I'm closing this in favor of that now. Again, sorry for not getting to this earlier!

Issue #444, opened 2024-07-17
Support `Transform` interpolation


janhohenheim:
- Should probably implement #263 first
- Closely related to https://github.com/bevyengine/bevy/issues/1259
- Take note of the fact that `GlobalTransform` is currently not updated between `FixedUpdate` calls: https://github.com/bevyengine/bevy/issues/7836
	- Should not be that important since it's a visual component and should be updated during the visual part of the frame

janhohenheim:
See my crate [`avian_interpolation`](https://github.com/janhohenheim/avian_interpolation) for a suggestion for how to handle this. It's not fit for being directly upstreamed as it completely removes the transform sync, which would be a much more controversial change, but it should be a good starting point. Another option is to just use [`bevy_transform_interpolation`](https://github.com/Jondolf/bevy_transform_interpolation) and accept the fact that it necessarily stores redundant data.

PR #512, opened 2024-09-10
Allow interpolation of scales

janhohenheim:
# Objective

For [`avian_interpolation`](https://github.com/janhohenheim/avian_interpolation), I want to have a clean separation between gameplay transforms and render transforms. Avian's best version of a gameplay scale is currently `Collider::scale`. However, this is not ergonomically interpolatable due to two reasons:
- `Collider::scale` is not initialized with the right scale, meaning that my observers that listen for its insertion will read wrong initial data, and
- It is automatically synced to the `GlobalTransform`, making it hard to change the `Transform` according to the `Collider::scale` without affecting it back in turn

Related to #444 

## Solution

- Set the `Collider::scale` correctly in its `on_add` hook
- Allow an opt-out via `SyncConfig`

---

## Changelog

- You can now opt-out of `Transform` to `Collider::scale` syncing by using the new `SyncConfig::transform_to_collider_scale` field.

### DIFF
--- a/src/collision/collider/backend.rs
+++ b/src/collision/collider/backend.rs
@@ -276,7 +276,9 @@ impl<C: ScalableCollider> Plugin for ColliderBackendPlugin<C> {
                     .in_set(PrepareSet::InitTransforms)
                     .after(init_transforms::<RigidBody>),
                 (
-                    update_collider_scale::<C>,
+                    update_collider_scale::<C>.run_if(|sync_config: Res<SyncConfig>| {
+                        sync_config.transform_to_collider_scale
+                    }),
                     update_collider_mass_properties::<C>,
                 )
                     .chain()
@@ -656,12 +658,7 @@ pub fn update_collider_scale<C: ScalableCollider>(
         // Child colliders
         Query<(&ColliderTransform, &mut C), (With<Parent>, Changed<ColliderTransform>)>,
     )>,
-    sync_config: Res<SyncConfig>,
 ) {
-    if !sync_config.transform_to_collider_scale {
-        return;
-    }
-
     // Update collider scale for root bodies
     for (transform, mut collider) in &mut colliders.p0() {
         #[cfg(feature = "2d")]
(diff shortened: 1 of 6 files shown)


janhohenheim:
@Jondolf could you help me out with the CI failure? With the `.into` it says that it is not needed for some features, but without it, we get this:
image
ofc I could just ignore the clippy lint, but I'd like to know how the rest of the Avian codebase does this. This is not a case of `.adjust_precision`, is it?

janhohenheim:
Yep, you were right :D

src/collision/collider/backend.rs
```diff
@@ -95,9 +95,53 @@ impl<C: ScalableCollider> Plugin for ColliderBackendPlugin<C> {
...
+                .copied()
+                .unwrap_or_default();
+            let global_transform = if existing_global_transform != GlobalTransform::IDENTITY {
+                // This collider was built deferred, probably via `ColliderConstructor`.
+                existing_global_transform
+            } else {
+                // This collider *may* have been `spawn`ed directly on a new entity.
+                // As such, its global transform is not yet available.
+                // You may notice that this will fail if the hierarchy's scale was updated in this
+                // frame. Remember that `GlobalTransform` is not updated inbetween fixed updates.
```
Jondolf, line 110:
```suggestion
                // frame. Remember that `GlobalTransform` is not updated in between fixed updates.
```

src/collision/collider/backend.rs
```diff
@@ -95,9 +95,53 @@ impl<C: ScalableCollider> Plugin for ColliderBackendPlugin<C> {
...
+                // But this is fine, as `update_collider_scale` will be updated in the next fixed update anyways.
```
Jondolf, line 111:
```suggestion
                // But this is fine, as `update_collider_scale` will be updated in the next fixed update anyway.
```

src/collision/collider/backend.rs
```diff
@@ -95,9 +95,53 @@ impl<C: ScalableCollider> Plugin for ColliderBackendPlugin<C> {
...
+                    .copied()
+                    .unwrap_or_default();
+                parent_global_transform * transform
+            };
+
+            let scale = global_transform.compute_transform().scale;
+            #[cfg(feature = "2d")]
+            let scale = scale.xy();
+
+            // Make sure the is initialized with the correct scale.
```
Jondolf, line 133:
```suggestion
            // Make sure the collider is initialized with the correct scale.
```

Jondolf (approved):
Looks good, thanks!

PR #519, opened 2024-09-22
Make disabling joints possible.

shanecelis:
# Objective

Make disabling joints possible.

## Solution

I added `DisableJoint` and added the `Without<DisableJoint>` filter on the query.

### Naming

In discord Jondolf had used the name `JointDisabled`; however, I thought the verb form `DisableJoint` was my preference. But this is not a strong opinion. If there are other past tense components, then I'm happy to have it either way.

### Other Considerations

I added derives for copy, clone, reflect, and serialize but these were added to follow suit rather than for a good reason of my own. 

## Alternatives

One could set the compliance to a large number. But what large number? And why bother computing when one would prefer disabling?

---

## Changelog

- Joints can be disabled with the `DisableJoint` component.

### DIFF
--- a/src/dynamics/solver/joints/mod.rs
+++ b/src/dynamics/solver/joints/mod.rs
@@ -340,5 +340,8 @@ impl AngleLimit {
 }
 
 /// Disables the joint of the entity it is placed on.
-#[derive(Component, Debug)]
+#[derive(Reflect, Clone, Copy, Component, Debug)]
+#[cfg_attr(feature = "serialize", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serialize", reflect(Serialize, Deserialize))]
+#[reflect(Debug, Component)]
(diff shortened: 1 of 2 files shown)


Jondolf:
I think in this case, I'd use `JointDisabled`. It's a marker indicating the state of the joint, and for that, a descriptive word like `Disabled` might be better than the present tense verb `Disable`. Using the present verb here would imply to me that it is actively doing something, like queuing the joint to get disabled, rather than just saying "this joint is disabled".

If we're looking for examples of this in Bevy and other ECSs, Flecs has a [`Disabled`](https://www.flecs.dev/flecs/md_docs_2EntitiesComponents.html#disabling) tag, and Bevy will likely add one as well (an earlier PR for it is [here](https://github.com/bevyengine/bevy/pull/12928)).

Jondolf (approved):
Thanks!

PR #527, opened 2024-10-04
AnyCollider context

oxkitsune:
# Objective

Provide a way for the methods in `AnyCollider` to obtain extra context from the ecs/entity it is attached to.

## Solution

- Add a `Context` GAT to the `AnyCollider` trait
- Optionally query for this context in the broad/narrow phase
- Pass the `Option<Context>` to the `AnyCollider` methods when called

---

## Changelog

- Changed: `AnyCollider` requires a `Context` GAT, to provide additional context from the ecs.

### DIFF
--- a/src/collision/collider/backend.rs
+++ b/src/collision/collider/backend.rs
@@ -145,7 +145,7 @@ impl<C: ScalableCollider> Plugin for ColliderBackendPlugin<C> {
             let aabb = entity_ref
                 .get::<ColliderAabb>()
                 .copied()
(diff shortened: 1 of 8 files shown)


oxkitsune:
This is a draft, as I'm looking for comments/feedback on the implementation. 
Perhaps there is a better approach, but I'm not super familiar with avian's internals. Or this not something avian is looking for. either way, I'm curious to hear what you think 😄

Jondolf:
Adopted in #665, which was just merged. Closing :)

Issue #436, opened 2024-07-17
Add `RigidBodyDisabled` and `ColliderDisabled`


janhohenheim:
Needed by @kaosat-dev for nicer Blenvy support.

PR #536, opened 2024-10-17
Add `RigidBodyDisabled`

Jondolf:
# Objective

Fixes part of #436.

A pretty common user request is to be able to temporarily disable rigid bodies without removing them from the world. This can be done by removing `RigidBody`, but this loses the type of rigid body, makes queries a bit more annoying, and can involve somewhat expensive clean-up.

In the absence of general-purpose component disabling in Bevy, it would be useful to have a `RigidBodyDisabled` marker that can be used to temporarily remove a body from the simulation.

## Solution

Add a `RigidBodyDisabled` marker component. It disables velocity, forces, contact response, and attached joints. It does *not* disable collision detection or spatial queries for attached colliders however; that should be handled with a separate `ColliderDisabled` component in a follow-up.

### DIFF
--- a/src/dynamics/solver/mod.rs
+++ b/src/dynamics/solver/mod.rs
@@ -151,7 +151,7 @@ impl Plugin for SolverPlugin {
         // Solve velocities using a position bias.
         substeps.add_systems(
             (
-                |mut bodies: Query<RigidBodyQuery>,
+                |mut bodies: Query<RigidBodyQuery, Without<RigidBodyDisabled>>,
                  mut constraints: ResMut<ContactConstraints>,
                  solver_config: Res<SolverConfig>,
                  length_unit: Res<PhysicsLengthUnit>,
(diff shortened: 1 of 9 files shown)


src/dynamics/solver/mod.rs
```diff
@@ -149,57 +149,24 @@ impl Plugin for SolverPlugin {
...
-                        time.delta_seconds_adjusted(),
-                        1,
-                        true,
-                        solver_config.max_overlap_solve_speed * length_unit.0,
-                    );
-                },
-            )
-                .in_set(SubstepSolverSet::SolveConstraints),
-        );
+        substeps.add_systems(solve_contacts::<true>.in_set(SubstepSolverSet::SolveConstraints));
```
Jondolf, line 152: This is an unrelated change, but I also finally decided to clean up the system registration here and made it use named systems instead of the mysterious anonymous closures that show up in traces.

PR #538, opened 2024-10-23
Implement `Default` for `ColliderConstructor` for `avian2d`

Lubba-64:
# Objective

- Fixes the issue in the header.

## Solution

- Make a unit circle option to satisfy the compiler and select it as `default`

Why? Because `space_editor` requires components to implement `Default`, and I think this is just a good thing for 2d to be able to do.


Lubba-64:
I have made a pull request because I am genuinely struggling to "build" this repository locally, but I am quite confident this implementation will fix my issue.

Lubba-64:
oops meant to make a separate branch my bad.

PR #542, opened 2024-11-03
Fix 3D friction

### LINKED ISSUE #489: 3D friction behaves unrealistically
The current 3D friction implementation seems to be incorrect. As can be seen in the Gif in the README's "Usage example", the box stays essentially at the same spot despite dropping onto a corner.

Box dropping onto a platform

Intuitively, you would expect the box to tumble in some direction based on which corner hit the ground and how. In fact, this is what happens with a similar setup using Rapier.

Notably, our current behavior can be roughly replicated in Rapier with a friction of zero. It almost looks like our friction only kicks in once the box is flat on the ground and the contact points are clearly sliding along the surface.

A similar setup in 2D does seem to work correctly.

Jondolf:
# Objective

Fixes #489.

Friction in 3D currently behaves very unrealistically. In the README's example, the cube should bounce to the side as it lands on a corner:

Broken friction

Instead, the cube bounces up a bit, spins in the air, and falls in the middle of the platform.

And in a case with locked rotation and zero restitution, friction also causes a weird upward bounce:

https://github.com/user-attachments/assets/0eec421c-741e-4639-b74e-ee3c16cc6cc5

The culprit here is that the tangent directions for friction are computed wrong in 3D. For the initial contact, one of them is pointing along the contact normal, while the other is close to zero.

Incorrect tangent directions

## Solution

Fix the tangent directions. This turned out to be just a dumb sign error...

This changes 3D behavior quite a bit, but it's for the better. Friction and bounces are more realistic, and stacking is more stable. Bounces behave correctly:

3d_scene

(note, the initial angular velocity is slightly different here than in the earlier example, it looks better for the README)


tim-blackbird (approved):
Tested the before and after in the `cubes` example. The results now are much closer to rapier and what I'd expect generally :)

PR #547, opened 2024-11-04
Fix overlap warnings: yeet edition

NiseVoid:
# Objective

- Overlap warnings are annoying

## Solution

- Yeet

PR #548, opened 2024-11-04
Fix overlap warnings: the depth check edition

NiseVoid:
# Objective

- Overlap warnings are annoying

## Solution

- Only show them when the overlap is significant


Jondolf:
Closing in favor of yeet #547

PR #568, opened 2024-11-28
Fix non-square `Collider::heightfield` row and column count order

Jondolf:
# Objective

The row and column counts in `Collider::heightfield` seem to be the wrong way around! This makes heightfields with a different number of rows and columns behave incorrectly in comparison to the documented representation.

The heightfield below should slope up smoothly left to right.

broken

## Solution

Fix the order.

fixed

### DIFF
--- a/src/collision/collider/parry/mod.rs
+++ b/src/collision/collider/parry/mod.rs
@@ -868,7 +868,7 @@ impl Collider {
             "Each row in `heights` must have the same amount of points"
         );
 
-        let heights = nalgebra::DMatrix::from_vec(row_count, column_count, data);
+        let heights = nalgebra::DMatrix::from_vec(column_count, row_count, data);
         SharedShape::heightfield(heights, scale.into()).into()
     }
 


src/collision/collider/parry/mod.rs
```diff
@@ -868,7 +868,7 @@ impl Collider {
             "Each row in `heights` must have the same amount of points"
         );
 
-        let heights = nalgebra::DMatrix::from_vec(row_count, column_count, data);
+        let heights = nalgebra::DMatrix::from_vec(column_count, row_count, data);
```
Nopey, line 871: Do not pass `column_count` to the `nrows` argument.

DMatrix::from_vec's first argument is `nrows`, and second is `ncol`; the bug is in the calculation of column and row count.
Instead of switching the order of the arguments, the two variables' assignments should be switched.

Nopey (changes requested):
`DMatrix::from_vec` expects column-major order.

`Collider::heightfield` does not currently document whether `heights` is a vector of column vectors (equivalently, that it is column major) or (if a transpose is added in heightfield) that it is a vector of row vectors (equiv. row major), could this PR also address that?

Thanks, and I hope this review is welcome.

PR #597, opened 2024-12-16
Run `update_collider_parents` before `update_root_collider_parents` to resolve ordering ambiguity on `ColliderParent`

SirHall:
…

# Objective
Fixes issue with system ordering where `update_collider_parents` and `update_root_collider_parents` both mutably access `ColliderParent` at the same time.

## Solution

I have ordered `update_root_collider_parents` to run `update_collider_parents` but feel free to modify if there's a better approach.


Jondolf:
#671 reworked this logic to use observers, so there shouldn't be any ambiguities or other scheduling problems anymore.

Sorry for not responding to this earlier!

PR #606, opened 2024-12-25
Draft: Collide and Slide Kinematic Character Controller Example (3d)

UndefinedBHVR:
# Objective

- Implements a collide and slide character controller based on the implementation I wrote for my personal project, Vidar.

## Solution

- Adds a new example, Collide and Slide 3d. Implements the a kinematic character controller using the algorithm from Kasper Fauerby + my own personal research in changes.
- The character controller is a multipass resolver, gravity is handled separately to ensure actual movement logic doesn't fight with it. For some games, this is preferable, for others not so much. We may want to document that.
---

## Changelog

- Implemented Collide and Slide 3d character controller example.

## Migration Guide

> This section is optional. If there are no breaking changes, you can delete this section.

- No actions required.

### DIFF
--- a/crates/avian3d/Cargo.toml
+++ b/crates/avian3d/Cargo.toml
@@ -101,6 +101,10 @@ name = "kinematic_character_3d"
 required-features = ["3d", "default-collider", "bevy_scene"]
 
 [[example]]
+name = "collide_and_slide_3d"
+required-features = ["3d", "default-collider", "bevy_scene"]
+
+[[example]]
 name = "cast_ray_predicate"
 required-features = ["3d", "default-collider"]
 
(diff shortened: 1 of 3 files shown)


fr33zing:
Hi, I'm working on a KCC and I've used some parts of yours. I noticed an issue where any slope is climbable. I checked your Vidar code and it seems like this issue would be present there too.

I have a solution to this, here's a snippet from mine:
```rs
let too_steep = hit.normal1.angle_between(Vec3::Y) > MAX_SLOPE_DEGREES.to_radians();
let normal = if pass == Pass::Movement && too_steep {
    (hit.normal1 * Vec3::new(1.0, 0.0, 1.0)).normalize()
} else {
    hit.normal1
};
```
Then `normal` goes into `calculate_sliding_velocity`. This makes steep slopes act like walls. I'm curious if you'd consider this a proper solution or not.

PR #616, opened 2025-01-05
Improved contact types

Jondolf:
# Objective

A lot of Avian's current contact types store some redundant data, aren't properly optimized, and aren't explicit enough about what they represent.

This PR is a clean-up pass to make the types a bit smaller and clearer.

### `Contacts`

- Replace `total_normal_impulse` property with `total_normal_impulse`, `total_normal_impulse_magnitude`, and `max_normal_impulse` helpers
- Deprecate `total_normal_force` helper
  - You can just divide by the substep timestep
- Remove `total_tangent_impulse` property and `total_friction_force` helper
  - These could previously technically be wrong/misleading, since each contact point can have a different tangent direction, especially in 3D, and they can also change between substeps. The tangent impulse magnitudes of each individual point can still be accessed.

### `ContactManifold`

- Rename `ContactManifold::contacts` to `ContactManifold::points`
- Use an `ArrayVec` with a capacity of 2 instead of a `Vec` to store 2D `ContactManifold` points directly on the stack
- Replace the local `normal1` and `normal2` with a single world-space `normal`
  - Internals only need the world-space normal, and it makes e.g. contact modification more straightforward
- Add `total_normal_impulse` and `max_normal_impulse` helpers

### `ContactData`

- Rename `ContactData` to `ContactPoint` (represents a point in a contact manifold)
  - Another option would be `ManifoldPoint` like in Box2D; not sure which one is better
- Rename `point1` and `point2` to `local_point1` and `local_point2` for explicitness
- Remove `normal1` and `normal2` from `ContactPoint`, since the normal is already stored in the `ContactManifold`

### Other

- Many documentation improvements
- Contact force debug rendering uses per-point forces instead of the total force for the contact pair

---

## Migration Guide

There have been several changes to Avian's contact types to make them more optimized and clearer.

### `Contacts`

- The `total_normal_impulse` property has been replaced with a `total_normal_impulse` helper method.
- The `total_normal_force` helper has been deprecated. Instead, just divide the impulse by the substep timestep.
- The `total_tangent_impulse` property and `total_friction_force` helper have been removed for being inaccurate/misleading. The tangent impulse magnitudes of each individual point can still be accessed.

### `ContactManifold`

- `ContactManifold::contacts` has been renamed to `ContactManifold::points`.
- The local `normal1` and `normal2` have been replaced with a single world-space `normal`, pointing from the first shape to the second.

### `ContactData`

- `ContactData` has been renamed to `ContactPoint`, since it specifically represents a point in a contact manifold, not general contact data.
- `point1` and `point2` have been renamed to `local_point1` and `local_point2` for explicitness
- `normal1` and `normal2` have been removed, since the normal is already stored in the `ContactManifold`.

### DIFF
--- a/src/collision/mod.rs
+++ b/src/collision/mod.rs
@@ -342,27 +342,15 @@ impl Contacts {
 pub struct ContactManifold {
     /// The contacts in this manifold.
     pub contacts: Vec<ContactData>,
-    /// A contact normal shared by all contacts in this manifold,
-    /// expressed in the local space of the first entity.
-    pub normal1: Vector,
-    /// A contact normal shared by all contacts in this manifold,
-    /// expressed in the local space of the second entity.
-    pub normal2: Vector,
+    /// The unit contact normal in world space, pointing from the first entity to the second.
+    ///
+    /// The same normal is shared by all contact points in this manifold,
+    pub normal: Vector,
     /// The index of the manifold in the collision.
     pub index: usize,
 }
 
 impl ContactManifold {
-    /// Returns the world-space contact normal pointing towards the exterior of the first entity.
-    pub fn global_normal1(&self, rotation: &Rotation) -> Vector {
-        rotation * self.normal1
-    }
-
-    /// Returns the world-space contact normal pointing towards the exterior of the second entity.
-    pub fn global_normal2(&self, rotation: &Rotation) -> Vector {
-        rotation * self.normal2
-    }
-
     /// Copies impulses from previous contacts to matching contacts in `self`.
     ///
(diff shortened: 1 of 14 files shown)


src/collision/mod.rs
```diff
@@ -260,51 +260,59 @@ pub struct Contacts {
     pub during_current_frame: bool,
     /// True if the bodies were in contact during the previous frame.
     pub during_previous_frame: bool,
```
Jondolf, line 262: These and `is_sensor` should use a `ContactPairFlags` bitflag type or similar, but I'll do that in another PR where I rework contact pair management.

PR #627, opened 2025-01-13
Fix use of unmaintaned crate proc-macro-error

PhantomMorrigan:
# Objective

The Avian macro crate uses `proc-macro-error`, which is [unmaintained](https://rustsec.org/advisories/RUSTSEC-2024-0370), this also leads to an outdated version of `syn` in the dependency tree.

## Solution

Replace `proc-macro-error` with `proc-macro-error2`.


Jondolf (approved):
Thanks! I'll publish a patch release for `avian_derive` in a follow-up PR.

PR #630, opened 2025-01-14
Make avian3d::dynamics::rigid_body::forces public

### LINKED ISSUE #626: avian3d::dynamics::rigid_body::forces is only exported via prelude
see: https://github.com/bevyengine/bevy/issues/17337#issuecomment-2586239119

SetOfAllSets:
# Objective

Fixes #626

## Solution

Make avian3d::dynamics::rigid_body::forces public

---

## Changelog

avian3d::dynamics::rigid_body::forces is now public


Jondolf (approved):
Thanks, good catch!

I added some basic module-level docs since CI was complaining about missing docs for `forces` now that it's public.

Issue #631, opened 2025-01-15
Replace `RenderLayers` in `PhysicsPickingPlugin` with a filter for `CollisionLayers`


Jondolf:
The `PhysicsPickingPlugin` requires that the `RenderLayers` of cameras and pickable colliders match, if specified. Using `RenderLayers` for mesh and sprite picking makes sense to ensure that what you see is what you get, but for colliders, there is no connection to rendering or visuals.

We should instead provide a component like `PhysicsPickingLayers` or `PhysicsPickingFilter` to filter pickable entities by `CollisionLayers`. Alternatively, we could have a filter property on the `PhysicsPickingSettings` resource, but it wouldn't allow per-camera configuration.

## Additional Context

The original [`AvianBackend`](https://github.com/aevyrie/bevy_mod_picking/blob/7494b9614a50a9c25e77a9bdccdb9c010a506ae9/backends/bevy_picking_avian/src/lib.rs) in `bevy_mod_picking` used `RenderLayers`, which is where the current implementation originates from.

Jondolf:
Fixed by #632

PR #632, opened 2025-01-15
Introduce `PhysicsPickingFilter`

morgenthum:
See #631


morgenthum:
I removed the `RenderLayers` checks, because it would be too complicated to configure picking, if we have `CollisionLayers`, `RenderLayers` and the new `PhysicsPickingLayers` on the camera.

Jondolf (approved):
Thanks for the PR! I pushed some changes to replace `PhysicsPickingLayers` with a `PhysicsPickingFilter` component that stores a whole `SpatialQueryFilter` instead of just a layer mask, like I [proposed on Discord](https://discord.com/channels/691052431525675048/1124043933886976171/1333121685796425870) :)

Also made some small doc improvements and added the component to the prelude. I intend to merge this tomorrow.

PR #638, opened 2025-01-22
Fix SpatialPipeline with_predicate docs

cBournhonesque:
# Objective
The docs say that the ray/shape casts will stop on the first instance where the predicate returns false, but it seems like the opposite is the case.

Maybe it would be good to add a unit test to confirm this.

### DIFF
--- a/src/spatial_query/pipeline.rs
+++ b/src/spatial_query/pipeline.rs
@@ -186,7 +186,7 @@ impl SpatialQueryPipeline {
     /// - `max_distance`: The maximum distance the ray can travel.
     /// - `solid`: If true *and* the ray origin is inside of a collider, the hit point will be the ray origin itself.
     ///   Otherwise, the collider will be treated as hollow, and the hit point will be at its boundary.
-    /// - `predicate`: A function called on each entity hit by the ray. The ray keeps travelling until the predicate returns `false`.
+    /// - `predicate`: A function called on each entity hit by the ray. The ray keeps travelling until the predicate returns `true`.
     /// - `filter`: A [`SpatialQueryFilter`] that determines which colliders are taken into account in the query.
     ///
     /// # Related Methods
@@ -370,7 +370,7 @@ impl SpatialQueryPipeline {
     /// - `direction`: What direction the shape is cast in.
     /// - `config`: A [`ShapeCastConfig`] that determines the behavior of the cast.
     /// - `filter`: A [`SpatialQueryFilter`] that determines which colliders are taken into account in the query.
-    /// - `predicate`: A function called on each entity hit by the shape. The shape keeps travelling until the predicate returns `false`.
+    /// - `predicate`: A function called on each entity hit by the shape. The shape keeps travelling until the predicate returns `true`.
     ///
     /// # Related Methods
     ///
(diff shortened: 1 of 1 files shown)


src/spatial_query/pipeline.rs
```diff
@@ -474,7 +474,7 @@ impl SpatialQueryPipeline {
 
     /// Casts a [shape](spatial_query#shapecasting) with a given rotation and computes computes all [hits](ShapeHitData)
     /// in the order of distance, calling the given `callback` for each hit. The shapecast stops when
-    /// `callback` returns false or all hits have been found.
+    /// `callback` returns true or all hits have been found.
```
Jondolf, line 477: I'm pretty sure the old docs are correct here. The cast continues until `false` is returned by the callback, like Rapier's [`intersections_with_shape`](https://docs.rs/bevy_rapier3d/0.28.0/bevy_rapier3d/plugin/struct.RapierContext.html#method.intersections_with_shape).

src/spatial_query/pipeline.rs
```diff
@@ -640,7 +640,7 @@ impl SpatialQueryPipeline {
 
     /// An [intersection test](spatial_query#intersection-tests) that finds all entities with a [collider](Collider)
     /// that contains the given point, calling the given `callback` for each intersection.
-    /// The search stops when `callback` returns `false` or all intersections have been found.
+    /// The search stops when `callback` returns `true` or all intersections have been found.
```
Jondolf, line 643: Same here

src/spatial_query/pipeline.rs
```diff
@@ -692,7 +692,7 @@ impl SpatialQueryPipeline {
 
     /// An [intersection test](spatial_query#intersection-tests) that finds all entities with a [`ColliderAabb`]
     /// that is intersecting the given `aabb`, calling `callback` for each intersection.
-    /// The search stops when `callback` returns `false` or all intersections have been found.
+    /// The search stops when `callback` returns `true` or all intersections have been found.
```
Jondolf, line 695: And here

src/spatial_query/pipeline.rs
```diff
@@ -747,7 +747,7 @@ impl SpatialQueryPipeline {
 
     /// An [intersection test](spatial_query#intersection-tests) that finds all entities with a [`Collider`]
     /// that is intersecting the given `shape` with a given position and rotation, calling `callback` for each
-    /// intersection. The search stops when `callback` returns `false` or all intersections have been found.
+    /// intersection. The search stops when `callback` returns `true` or all intersections have been found.
```
Jondolf, line 750: ...And here

src/spatial_query/pipeline.rs
```diff
@@ -259,7 +259,7 @@ impl SpatialQueryPipeline {
     }
 
     /// Casts a [ray](spatial_query#raycasting) and computes all [hits](RayHitData), calling the given `callback`
-    /// for each hit. The raycast stops when `callback` returns false or all hits have been found.
+    /// for each hit. The raycast stops when `callback` returns true or all hits have been found.
```
Jondolf, line 262: This one too

Jondolf (commented):
Thanks, yeah I think the docs for the predicates are currently wrong. I'm pretty sure the old callback docs are correct though.

cBournhonesque:
> Thanks, yeah I think the docs for the predicates are currently wrong. I'm pretty sure the old callback docs are correct though.

Oh I see; I didn't test for callback but I assumed they were also inverted; will fix

Jondolf (approved):
Thanks!

PR #665, opened 2025-03-07
Add Context to AnyCollider

NiseVoid:
# Objective

- Updated version of #527 with some changes suggested by Jondolf

## Solution

- Reapply the changes from the previous PR
- Add a `SimpleCollider` trait that works for colliders with no context to shorten function signatures

## Changelog

- Custom colliders can now specify a SystemParam under Context
- The context is passed to all functions allowing data from the World to be fetched
- AnyCollider functions have been prefixed with `get_`
- A new `SimpleCollider` trait has been added to add functionality to colliders without context

## Migration Guide

- Custom colliders now need to specify a `Context` `SystemParam`, if this is unnecessary `()` should be used
- When trying to use methods on `AnyCollider` for the defualt collider or other colliders without context, `SimpleCollider` should now be used instead
- Methods on `AnyCollider` have been prefixed with `get_` to keep the name and usage of the simplified versions unchanged.


src/collision/collider/mod.rs
```diff
@@ -42,16 +45,167 @@ pub trait IntoCollider<C: AnyCollider> {
...
+
+impl<'a, 'w, 's, T: ReadOnlySystemParam> AabbContext<'a, 'w, 's, T> {
+    /// Construct an [`AabbContext`]
+    pub fn new(entity: Entity, item: &'a <T as SystemParam>::Item<'w, 's>) -> Self {
+        Self { entity, item }
+    }
+}
+
+impl AabbContext<'_, '_, '_, ()> {
+    fn fake() -> Self {
```
Jondolf, line 74: Maybe `empty` would be a better name? Though I guess `Entity::PLACEHOLDER` isn't necessarily "empty"

src/collision/collider/mod.rs
```diff
@@ -42,16 +45,167 @@ pub trait IntoCollider<C: AnyCollider> {
...
         feature = "2d",
         doc = "\n\nThe rotation is counterclockwise and in radians."
     )]
-    fn aabb(&self, position: Vector, rotation: impl Into<Rotation>) -> ColliderAabb;
+    fn aabb_with_context(
+        &self,
+        context: AabbContext<Self::Context>,
+        position: Vector,
+        rotation: impl Into<Rotation>,
+    ) -> ColliderAabb;
```
Jondolf, line 208: Personally this argument order still feels a bit weird to me. The parameters of the AABB are really the main thing here, and the context is just an extra thing to provide more contextual data. I think it'd be more natural to have the context be the last argument here and in the other `_with_context` methods.

src/collision/collider/mod.rs
```diff
@@ -60,24 +214,26 @@ pub trait AnyCollider: Component + ComputeMassProperties {
...
     /// Returns an empty vector if the colliders are separated by a distance greater than `prediction_distance`
     /// or if the given shapes are invalid.
-    fn contact_manifolds(
+    fn contact_manifolds_with_context(
         &self,
         other: &Self,
+        context: ContactManifoldContext<Self::Context>,
         position1: Vector,
         rotation1: impl Into<Rotation>,
         position2: Vector,
```
Jondolf, line 239: Same as the AABB methods, I feel like the context should be at the end. In this case I also consider the positions and rotations to be kinda "part of" the shape definition, so it's natural to have them next to each other, while the context isn't strictly related to the shapes themselves.

src/collision/collider/mod.rs
```diff
@@ -86,6 +242,41 @@ pub trait AnyCollider: Component + ComputeMassProperties {
     ) -> Vec<ContactManifold>;
 }
 
+/// A simplified wrapper around [`AnyCollider`] that doesn't require passing in the context for
+/// implementations that don't need context
+pub trait SimpleCollider: AnyCollider<Context = ()> {
+    /// [`AnyCollider::aabb_with_context`] but without context
+    fn aabb(&self, p: Vector, r: impl Into<Rotation>) -> ColliderAabb {
```
Jondolf, line 249: Given that these are the simpler versions of the methods, and what most people will probably reach for first unless specifically dealing with contexts, I think these should have proper docs like the methods on `AnyCollider`.

Could just copy them over and have some comment mentioning the context version, maybe something like this?

```suggestion
    /// Computes the [Axis-Aligned Bounding Box](ColliderAabb) of the collider
    /// with the given position and rotation.
    #[cfg_attr(
        feature = "2d",
        doc = "\n\nThe rotation is counterclockwise and in radians."
    )]
    ///
    /// For a version that provides access to an ECS context, see [`AnyCollider::aabb_with_context`].
    fn aabb(&self, p: Vector, r: impl Into<Rotation>) -> ColliderAabb {
```

Jondolf: I'd also use full names for the arguments, there's not really a reason to abbreviate here.

These methods could also be inlined, though it's probably done automatically anyway.

src/collision/collider/mod.rs
```diff
@@ -42,16 +45,167 @@ pub trait IntoCollider<C: AnyCollider> {
...
+    }
+}
+
+/// Context necessary to calculate [`ContactManifold`]s for a set of [`AnyCollider`]
+#[derive(Deref)]
+pub struct ContactManifoldContext<'a, 'w, 's, T: ReadOnlySystemParam> {
+    /// The first entity used in the contact manifold
+    pub entity1: Entity,
+    /// The second entity used in the contact manifold
+    pub entity2: Entity,
```
Jondolf, line 88:
```suggestion
    /// The first collider entity involved in the contact.
    pub entity1: Entity,
    /// The second collider entity involved in the contact.
    pub entity2: Entity,
```

Jondolf (commented):
Looks pretty good, mainly a few stylistic nits and documentation suggestions.

Just to be sure, I'll test later today if this has any measurable perf impact in the default case of an empty context, but I assume it should be basically zero-cost.

Jondolf (approved):
Looks good now! Thanks for the patience on this @NiseVoid and @oxkitsune, I've kept the collider context work sitting here for too long.

I made a couple more small changes just to clean things up, and also ran a quick test to see if this has any measurable perf impact. I didn't notice anything, so it should be good to go :)

PR #666, opened 2025-03-10
Add `CustomGravity` component

Pyxrs:
# Objective

Some users may want gravity to affect bodies differently, such as for Newtonian gravity or other nonstandard applications of Avian. `GravityScale` exists, but it is a very limited solution.

## Solution

I added a `CustomGravity` component that overrides the `Gravity` resource if it is present on a body.

---

## Changelog

- Added: `CustomGravity`
- Added: `custom_gravity` example (based on `many_shapes`)

### DIFF
--- a/crates/avian2d/examples/custom_gravity.rs
+++ b/crates/avian2d/examples/custom_gravity.rs
@@ -1,0 +1,141 @@
+#![allow(clippy::unnecessary_cast)]
+
+use avian2d::{dynamics::integrator::CustomGravity, math::*, prelude::*};
+use bevy::prelude::*;
+use examples_common_2d::ExampleCommonPlugin;
+
+fn main() {
+    App::new()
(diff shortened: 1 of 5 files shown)


Pyxrs:
Looks like fmt messed up the diff, the only changes to the integrator module were on updated lines 210 and 244.

Jondolf:
Hi, thanks for the PR, and sorry for not responding earlier :sweat_smile:

I think this is a bit too niche in terms of the scope; fundamentally this is about a component to apply a constant linear acceleration to bodies. This doesn't necessarily need to override gravity (you could set `GravityScale` to zero if you did want that).

I just merged #770, which overhauled our force APIs and integration logic, and also included a `ConstantLinearAcceleration` component. It seems like that component fills the same need as this PR while being a bit more general. So I think I'll close this in favor of that :)

PR #668, opened 2025-03-21
Refactor to minimise `std` usage

bushrat011899:
# Objective

- Prepare for possible `no_std` support

## Solution

- Added lints and switched to using `core` and `alloc` instead of `std`.

### DIFF
--- a/crates/avian2d/Cargo.toml
+++ b/crates/avian2d/Cargo.toml
@@ -95,6 +95,11 @@ bytemuck = "1.19"
 criterion = { version = "0.5", features = ["html_reports"] }
 bevy_mod_debugdump = { version = "0.12" }
 
+[lints.clippy]
+alloc_instead_of_core = "warn"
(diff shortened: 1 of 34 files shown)


crates/avian2d/Cargo.toml
```diff
@@ -95,6 +95,11 @@ bytemuck = "1.19"
 criterion = { version = "0.5", features = ["html_reports"] }
 bevy_mod_debugdump = { version = "0.12" }
 
+[lints.clippy]
+alloc_instead_of_core = "warn"
+std_instead_of_alloc = "warn"
+std_instead_of_core = "warn"
+
```
Jondolf, line 102: Just a thought, can these be workspace lints so we don't need them in every `Cargo.toml`?

bushrat011899: Oh yeah definitely

Jondolf (approved):

PR #674, opened 2025-03-22
fix 2d examples not compiling

spookyvision:
# Objective

- fix 2d examples not compiling, error was: 
```
$ cargo run --example $ANY_AVIAN2D_EXAMPLE
error[E0433]: failed to resolve: use of undeclared type `Animatable`
   --> crates/avian2d/../../src/diagnostics/ui.rs:434:24
    |
434 |         text_color.0 = Animatable::interpolate(&green, &red, t.min(1.0)).into();
    |                        ^^^^^^^^^^ use of undeclared type `Animatable`
```

## Solution

- add `bevy_animation` to bevy feature flags in `examples_common_2d`

### DIFF
--- a/crates/examples_common_2d/Cargo.toml
+++ b/crates/examples_common_2d/Cargo.toml
@@ -8,6 +8,7 @@ use-debug-plugin = []
 
 [dependencies]
 bevy = { version = "0.16.0-rc.1", default-features = false, features = [
+    "bevy_animation",
     "bevy_core_pipeline",
     "bevy_state",
     "bevy_text",


Jondolf:
Hi, thanks and sorry for getting to this so late!

There's still a few issues here: I believe this only fixes the problem for the 2D examples, not 3D or for Avian itself. If someone was using Avian without having `bevy_animation` enabled for Bevy, they could get the same problem.

I've fixed this in #711 by just getting rid of the use of `Animatable` altogether.

PR #694, opened 2025-04-12
Handle entity disabling properly

Jondolf:
# Objective

Bevy 0.16 introduces entity disabling via the `Disabled` component. We need to handle this for the contact graph and broad phase.

## Solution

Remove disabled entities from the contact graph using observers, and add re-enabled collider entities to the broad phase.

### DIFF
--- a/src/collision/broad_phase.rs
+++ b/src/collision/broad_phase.rs
@@ -8,7 +8,7 @@ use core::marker::PhantomData;
 use crate::{data_structures::pair_key::PairKey, prelude::*};
 use bevy::{
     ecs::{
-        entity::{EntityMapper, MapEntities},
+        entity::{EntityHashSet, EntityMapper, MapEntities},
         entity_disabling::Disabled,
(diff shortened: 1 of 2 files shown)


src/collision/broad_phase.rs
```diff
@@ -206,11 +207,16 @@ type AabbIntervalQueryData = (
...
     mut intervals: ResMut<AabbIntervals>,
     mut re_enabled_colliders: RemovedComponents<ColliderDisabled>,
+    mut re_enabled_entities: RemovedComponents<Disabled>,
 ) {
-    let re_enabled_aabbs = aabbs.iter_many(re_enabled_colliders.read());
+    let re_enabled_aabbs = aabbs.iter_many(
+        re_enabled_colliders
+            .read()
+            .chain(re_enabled_entities.read()),
+    );
```
Jondolf, line 219: This is gross, but we'll handle it better with observers or hooks once we rework the broad phase to use a BVH.

PR #697, opened 2025-04-13
Add support for floating origins using deltas in sync plugin

austincummings:
# Objective

- Update the `SyncPlugin` to be compatible with floating origin plugins like `big_space` 

## Solution

- Use position deltas in the `SyncPlugin` to update transform translation

---

Creating this as a draft to discuss the change since I'm not sure if I broke hierarchy's in this. All the demos seem to work though so...!

Alternatively I was thinking there could be another config option in `SyncConfig` like `use_delta_transforms` that would call a different `position_to_transform_with_deltas` to trigger this functionality.


austincummings:
Looks like this may not be needed with https://github.com/Jondolf/bevy_transform_interpolation/tree/bevy-0.15-smooth-teleports.

But figured I'd submit my solution anyway even if it's not the preferred way to do it. :)

PR #701, opened 2025-04-15
Updated to Bevy 0.16.0-rc.5

lufog:
# Objective

- Fixes build with Bevy 0.16.0-rc5

## Solution

Micro PR, just adapts to bevyengine/bevy#18813, ~and removes a couple of unused imports~.

PR #702, opened 2025-04-15
Add support for bevy 0.16.0 rc5

guyguy2001:
# Objective

Supports bevy 0.16.0 rc 5; I'm pretty sure it breaks comparability with the previous RC versions.

(Since Cargo.toml doesn't pin an rc version, you will probably need to run `Cargo clean && bevy run` to see changes)

I hope this is useful!

## Solution / Changelog

I had to change platform_support to platform as per https://github.com/bevyengine/bevy/pull/18813

---

## Tests

I ran the following, and it seems to have worked:

```
cargo run --example cubes --no-default-features --features "3d f64 parry-f64"
```

### DIFF
--- a/src/collision/collider/constructor.rs
+++ b/src/collision/collider/constructor.rs
@@ -1,5 +1,5 @@
 use crate::prelude::*;
-use bevy::{platform_support::collections::HashMap, prelude::*};
+use bevy::{platform::collections::HashMap, prelude::*};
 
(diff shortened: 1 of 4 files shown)


Jondolf:
Hi! #701 actually has the exact same changes but was made just a bit earlier, so I merged that one 😅 Thank you for the PR though!

guyguy2001:
Whoops, I only checked the branches on the repo, instead of the PRs 😅 

Just as well though! Glad to see it supported so quickly

PR #710, opened 2025-04-25
Temporary support for bevy 0.16

MushineLament:
Only providing the simplest support for Bevy 0.16


Jondolf:
Thanks! The README's version table also needs to be updated though, and I'd prefer not to enable `bevy_animation` unconditionally like this (also it's still broken in 3D since this only enables it for 2D). I made #711 as an alternative to this, which should have these fixed

PR #724, opened 2025-05-08
Add Ease implementations on Position/Rotation

cBournhonesque:
# Objective

- Add Ease implementation on Position/Rotation so that they work well with bevy easing functions.

### DIFF
--- a/src/position.rs
+++ b/src/position.rs
@@ -103,7 +103,7 @@ impl From<&GlobalTransform> for Position {
 impl Ease for Position {
     fn interpolating_curve_unbounded(start: Self, end: Self) -> impl Curve<Self> {
         FunctionCurve::new(Interval::UNIT, move |t| {
-            Position(Vec2::lerp(start.0, end.0, t))
+            Position(Vector::lerp(start.0, end.0, t))
         })
     }
 }
--- a/src/position.rs
+++ b/src/position.rs
@@ -102,7 +102,9 @@ impl From<&GlobalTransform> for Position {
 
 impl Ease for Position {
     fn interpolating_curve_unbounded(start: Self, end: Self) -> impl Curve<Self> {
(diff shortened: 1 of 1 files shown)


src/position.rs
```diff
@@ -100,6 +100,14 @@ impl From<&GlobalTransform> for Position {
     }
 }
 
+impl Ease for Position {
+    fn interpolating_curve_unbounded(start: Self, end: Self) -> impl Curve<Self> {
+        FunctionCurve::new(Interval::UNIT, move |t| {
+            Position(Vec2::lerp(start.0, end.0, t))
```
Jondolf, line 106:
```suggestion
            Position(Vector::lerp(start.0, end.0, t))
```

src/position.rs
```diff
@@ -100,6 +100,14 @@ impl From<&GlobalTransform> for Position {
     }
 }
 
+impl Ease for Position {
+    fn interpolating_curve_unbounded(start: Self, end: Self) -> impl Curve<Self> {
+        FunctionCurve::new(Interval::UNIT, move |t| {
+            Position(Vector::lerp(start.0, end.0, t))
```
Jondolf, line 106:
```suggestion
            Position(Vector::lerp(start.0, end.0, t as Scalar))
```

src/position.rs
```diff
@@ -667,6 +675,12 @@ impl core::ops::Mul<&mut Vector3> for &mut Rotation {
     }
 }
 
+impl Ease for Rotation {
+    fn interpolating_curve_unbounded(start: Self, end: Self) -> impl Curve<Self> {
+        FunctionCurve::new(Interval::UNIT, move |t| Rotation::slerp(start, end, t))
```
Jondolf, line 680:
```suggestion
        FunctionCurve::new(Interval::UNIT, move |t| Rotation::slerp(start, end, t as Scalar))
```

Jondolf (commented):
Whoops, some math types here were still wrong

PR #725, opened 2025-05-09
Replaced Avian transform syncs with 0.16 Bevy versions

ramirezmike:
# Objective

- Replacing the Avian transform propagation functions with the ones from the Bevy engine to leverage [0.16's faster transform propagation](https://bevyengine.org/news/bevy-0-16/#faster-transform-propagation).

## Solution
Replaced every instance of these systems

```
sync_simple_transforms_physics,
propagate_transforms_physics,
```
with these directly from Bevy

```
mark_dirty_trees,
propagate_parent_transforms,
sync_simple_transforms,
```
And also removed code related to sync_simple_transforms_physics and propagate_transforms_physics that are now no longer used.

I was running into a problem on my computer where having many hierarchies was causing my FPS to vary wildly, but switching to the Bevy propagation resolved it. 

Before (averaging ~20 fps with drops into the single digits)
image
After (averaging a more steady ~100fps)
image

Using this resolved the issue in my game. I also ran a bunch of the examples to compare and can't tell a difference before and after even on the determinism ones.

---

## Changelog

> replaced the Avian transform propagation functions with the ones from the Bevy engine 

## Example Used

This is the example I was using above

```rust
use bevy::{
    diagnostic::{DiagnosticsStore, FrameTimeDiagnosticsPlugin},
    prelude::*,
};

use avian3d::prelude::*;

fn main() {
    App::new()
        .add_plugins((
            DefaultPlugins,
            PhysicsPlugins::default(),
            FrameTimeDiagnosticsPlugin::default(),
        ))
        .add_systems(Startup, setup)
        .add_systems(Update, text_update_system)
        .run();
}

#[derive(Component)]
struct FpsText;

fn setup(
    mut commands: Commands,
    mut meshes: ResMut<Assets<Mesh>>,
    mut materials: ResMut<Assets<StandardMaterial>>,
    mut time: ResMut<Time<Physics>>,
) {
    commands.spawn((
        Camera3d::default(),
        Transform::from_xyz(3.0, 1.0, 3.0).looking_at(Vec3::new(0.0, -0.5, 0.0), Vec3::Y),
    ));

    let cube = meshes.add(Cuboid::new(0.5, 0.5, 0.5));

    const GOLDEN_ANGLE: f32 = 137.507_77;

    let mut hsla = Hsla::hsl(0.0, 1.0, 0.5);
    let count = 20;
    let nest_count = 100;

    for x in -count..count {
        for z in -count..count {
            let mut current_entity = commands
                .spawn((
                    Transform::from_translation(Vec3::new(x as f32, 0.0, z as f32)),
                    Visibility::Visible,
                ))
                .id();

            for _ in 0..nest_count {
                current_entity = commands
                    .spawn((Transform::default(), Visibility::Visible))
                    .insert(ChildOf(current_entity))
                    .id();
            }

            commands.spawn((
                Mesh3d(cube.clone()),
                MeshMaterial3d(materials.add(Color::from(hsla))),
                Transform::default(),
                Collider::cuboid(0.5, 0.5, 0.5),
                RigidBody::Dynamic,
                ChildOf(current_entity),
            ));
            hsla = hsla.rotate_hue(GOLDEN_ANGLE);
        }
    }

    time.pause();

    commands
        .spawn((
            Text::new("FPS: "),
            TextFont {
                font_size: 42.0,
                ..default()
            },
        ))
        .with_child((
            TextSpan::default(),
            TextFont {
                font_size: 33.0,
                ..default()
            },
            FpsText,
        ));
}

fn text_update_system(
    diagnostics: Res<DiagnosticsStore>,
    mut query: Query<&mut TextSpan, With<FpsText>>,
) {
    for mut span in &mut query {
        if let Some(fps) = diagnostics.get(&FrameTimeDiagnosticsPlugin::FPS) {
            if let Some(value) = fps.smoothed() {
                **span = format!("{value:.2}");
            }
        }
    }
}
```

### DIFF
--- a/src/collision/collider/collider_transform/plugin.rs
+++ b/src/collision/collider/collider_transform/plugin.rs
@@ -7,6 +7,7 @@ use crate::{
 use bevy::{
     ecs::{intern::Interned, schedule::ScheduleLabel},
     prelude::*,
+    transform::systems::{mark_dirty_trees, propagate_parent_transforms, sync_simple_transforms},
 };
 
 /// A plugin for propagating and updating transforms for colliders.
@@ -51,8 +52,9 @@ impl Plugin for ColliderTransformPlugin {
         app.add_systems(
             self.schedule,
             (
-                crate::sync::sync_simple_transforms_physics,
-                crate::sync::propagate_transforms_physics,
+                mark_dirty_trees,
+                propagate_parent_transforms,
+                sync_simple_transforms,
             )
                 .chain()
                 .run_if(match_any::<(Added<ColliderMarker>, Without<RigidBody>)>)
--- a/src/prepare.rs
+++ b/src/prepare.rs
@@ -8,6 +8,7 @@ use crate::{prelude::*, sync::SyncConfig};
 use bevy::{
     ecs::{intern::Interned, query::QueryFilter, schedule::ScheduleLabel},
     prelude::*,
+    transform::systems::{mark_dirty_trees, propagate_parent_transforms, sync_simple_transforms},
 };
 
 /// Runs systems at the start of each physics frame. Initializes [rigid bodies](RigidBody)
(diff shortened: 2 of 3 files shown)


Jondolf (approved):
Thank you!

PR #729, opened 2025-05-12
Improve CI performance, cache size and reactivate Linux tests

janhohenheim:
- Create only new caches from `main`
- Reactivate Linux builds
  - Except for `collider_constructor_hierarchy_inserts_correct_configs_on_scene`, which requires the GPU for some reason
  - I thought this setup would be enough, but apparently it wasn't, since `SceneInstanceReady` is never triggered:
```rust
app.add_plugins((
    MinimalPlugins,
    AssetPlugin::default(),
    ScenePlugin,
    TransformPlugin,
    MeshPlugin,
    GltfPlugin::default(),
    ImagePlugin::default(),
    PhysicsPlugins::default(),
));
```
- Use nextest for tests
  - nextest cannot do doctests, so we still use `cargo test --doc` for those
- Use fixed cache keys against cache misses
- Share cache keys for lints to reduce cache size
- Remove some debug information from the builds to reduce cache size. No worries, the test error backtraces are not affected by this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,6 +52,9 @@ jobs:
               with:
                 save-if: ${{ github.ref == 'refs/heads/main' }}
                 shared-key: test-${{ matrix.os }}
+            - name: Run doc tests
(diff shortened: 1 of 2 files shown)


Jondolf (approved):
Thanks! Excited to see how this affects things

PR #743, opened 2025-05-26
Cache colliders created from meshes via constructors

janhohenheim:
# Objective

- Creating colliders from meshes can be expensive when not just using a trimesh
- Especially convex decomps take ages
- In end-user cases, `ColliderConstructor` and `ColliderConstructorHierarchy` are *very* popular for creating colliders from meshes
- Ideally, we would have a way to cache any and all colliders, but I think caching the ones made from a mesh with these methods would already bring us 90% of the way, from a user perspective.

## Solution

So, let's cache them! Intentionally leaving the cache `pub(crate)` and not `pub` so we can easily expand it without breaking anything for users.

---

## Changelog

- Using `ColliderConstructor` or `ColliderConstructorHierarchy` for creating colliders from meshes will now cache the collider. This is especially noticeable when creating multiple copies of a mesh with a collider from a convex decomposition, e.g. spawning ten chairs each with `ColliderConstructorHierarchy::new(ColliderConstructor::ConvexDecompositionFromMesh)`

### DIFF
--- a/src/collision/collider/backend.rs
+++ b/src/collision/collider/backend.rs
@@ -4,6 +4,8 @@
 
 use core::marker::PhantomData;
 
+#[cfg(feature = "collider-from-mesh")]
+use crate::collision::collider::cache::ColliderCache;
 use crate::{
     collision::broad_phase::BroadPhaseSet, prelude::*, prepare::PrepareSet, sync::SyncConfig,
 };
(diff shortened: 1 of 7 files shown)


Jondolf (approved):
Looks great, and the perf gains you shared on [Discord](https://discord.com/channels/691052431525675048/1124043933886976171/1376538700179308677) are impressive. Thanks!

PR #758, opened 2025-06-16
ColliderConstructor::Compound

Piefayth:
# Objective

Allow for the creation of compound colliders through `ColliderConstructor`.

## Solution

Added a new enum variant to `ColliderConstructor` and a helper impl to allow constructing the variant with a wider variety of types.

---

## Changelog

> Added `ColliderConstructor::Compound`


src/collision/collider/constructor.rs
```diff
@@ -280,6 +280,7 @@ pub struct ColliderConstructorHierarchyConfig {
 #[cfg_attr(feature = "collider-from-mesh", derive(Default))]
 #[cfg_attr(feature = "collider-from-mesh", reflect(Default))]
 #[reflect(Debug, Component, PartialEq)]
+#[reflect(no_field_bounds)]
```
janhohenheim, line 283: what does this do? :)

Issue #661, opened 2025-02-27
Customizable time dilation (timestep) for physics schedule


starwolfy:
Add the ability to set a custom delta time (timestep) that is used in `run_physics_schedule`. In a networked and rollbacked physics application you may run a subapp that runs the game schedules manually from an exclusive system in Bevy's `Main` schedule and it steps the precise amount of steps that it needs to perform to get the desired correction results. Correction subapp runs multiple ticks with the physics schedule still desiring to keep the delta time it calcutates consistent with the fixed delta time in series AND as fast as possible without actually waiting the fixed timestep amount. This is currently not possible without forking the engine. I do not think setting the relative speed helps in this case either.

It used to be possible in `bevy_xpbd` because the timestep would be determined from the Time<Physics> resource that could have a fixed timestep interval set, since the update to avian this has changed.

PR #817, opened 2025-09-16
Fix 1-frame pause delay

### LINKED ISSUE #776: `Time<Physics>` pause is delayed by one frame
From what I'm seeing, unpausing `Time<Physics>` takes effect immediately, but pausing it is delayed by one frame.

A simplified extract from my logs:
```text
# fixed schedule 1 (already paused)
after FixedPostUpdate, body position: [6.0856233, -7.507883] (same as before)
# fixed schedule 2 (already paused)
after FixedPostUpdate, body position: [6.0856233, -7.507883] (same as before)
# fixed schedule 3
on FixedUpdate, unpause Time<Physics>
after FixedPostUpdate, body position: [5.954998, -7.2947235] (updated)
# fixed schedule 4
on FixedUpdate, pause Time<Physics>
after FixedPostUpdate, body position: [5.824373, -7.081564] (updated)
# fixed schedule 5 (already paused)
after FixedPostUpdate, body position: [5.824373, -7.081564] (same as before)
```

I would expect the physics simulation to only run one time in this situation, but it seems to have run twice.

Pascualex:
Fixes #776. Details available on the issue.

Summary: currently if the user pauses `Time<Physics>` the pause is delayed by 1 frame because we detect the delta time of the previous frame as if the user had manually advanced the clock. This PR fixed the issue by always resetting the delta time. This shouldn't cause regressions since it doesn't interfere with the 2 standard ways to advance the simulation on the next frame: if the clock is not paused then the delta time will be automatically set at the beginning of `run_physics_schedule` and if the user manually advances the time then the reset will get overridden.

### DIFF
--- a/src/schedule/mod.rs
+++ b/src/schedule/mod.rs
@@ -253,13 +253,11 @@ fn run_physics_schedule(world: &mut World, mut is_first_run: Local<IsFirstRun>) 
             schedule.run(world);
         }
 
-        // If physics is paused, reset delta time to stop the simulation
-        // unless users manually advance `Time<Physics>`.
-        if is_paused {
-            world
-                .resource_mut::<Time<Physics>>()
-                .advance_by(Duration::ZERO);
-        }
+        // Reset delta time, to continue the simulation `Time<Physics>`
+        // must be unpaused or manually advanced by users.
+        world
+            .resource_mut::<Time<Physics>>()
+            .advance_by(Duration::ZERO);
 
         // Set the generic `Time` resource back to the clock that was active before physics.
         *world.resource_mut::<Time>() = old_clock;


Jondolf:
The delay is something we should fix, but the downside with this approach is that reading `Time<Physics>` in any schedule other than `PhysicsSchedule` will return a delta time of zero. While accessing it like this is likely rare in practice, it feels wrong that from an outside perspective, it looks like no time has passed for physics. Normally in Bevy, you can read clocks like `Time<Fixed>` in *any* schedule, and get the expected result.

Perhaps the fix would instead be to simply reset delta time when physics is paused (like before), and to *only* advance the simulation when it is not paused. This breaks the old supported approach of manually advancing `Time<Physics>` and having the `PhysicsSchedule` be run automatically, but I think this may be fine; manual stepping scenarios can (and arguably should) simply run `PhysicsSchedule` manually, instead of having it run whenever delta time is non-zero.

janhohenheim:
Related: #661

PR #819, opened 2025-09-19
Fix temporaries in bevy-0.17 branch

ChristopherBiscardi:
fix "temporaries freed while still in use" across a couple of function uses.

## Solution

I took the approach of keeping the clones of things like `DebugName` from the `.name` function, etc and instead stored the value so it could be referenced, since the usage relies on deref coercion to go from `DebugName` to `&str`.

The examples run after this fix


Jondolf (approved):
Thanks!

PR #869, opened 2025-10-20
Fix issue with `RayCaster` when max_hits is set to 1

### LINKED ISSUE #868: `RayCaster` does not populate `RayHits` when `max_hits` is set to 1
When `RayCaster` is used with `max_hits` set to 1, the `RayHits` component never gets populated with the hit results. It's a simple fix - see my PR :)

It would be great if this could be released as a patch to 0.4!

LlamaLamp:
# Objective

Fixes #868 

## Solution

Make sure the ray cast result is pushed into the `RayHits` component.

## Testing

I tested this locally against my own project, and it fixed the issue.

You can test this yourself by spawning a `RayCaster` with max_hits set to 1, and ensuring that the `RayHits` component gets updated with the result.

Are there unit tests or similar that I can add a test case to?

### DIFF
--- a/src/spatial_query/ray_caster.rs
+++ b/src/spatial_query/ray_caster.rs
@@ -262,13 +262,17 @@ impl RayCaster {
         hits.clear();
 
         if self.max_hits == 1 {
-            query_pipeline.cast_ray(
+            let first_hit = query_pipeline.cast_ray(
                 self.global_origin(),
                 self.global_direction(),
                 self.max_distance,
                 self.solid,
                 &self.query_filter,
             );
+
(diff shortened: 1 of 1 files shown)


Jondolf (approved):
Oof, not sure how that bug happened 😬 Looks like it was introduced by #815, and I somehow didn't realize that it wasn't actually pushing the result.

I'll most likely do a 0.4.1 patch release this weekend; there's a few other small bugs that would be good to fix too.

We don't have good automated tests for spatial queries yet, though it would be good to have them. But we don't need to do that in this PR, I opened #871 to track it

PR #889, opened 2025-11-13
Make normals `Dir`s

janhohenheim:
# Objective

- Normals should be normalized
- But our type system says they may not be!

## Solution

- Make them `Dir`


janhohenheim:
There's no `Dir` for `f64`, heck

PR #894, opened 2025-11-17
Implement move-and-slide

janhohenheim:
# Objective

- Resolves https://github.com/avianphysics/avian/issues/438
- Supersedes https://github.com/avianphysics/avian/pull/606
- Almost all kinematic character controllers use an algorithm called "move and slide", aka "collide and slide", "step slide", and similar, at their core
- This algorithm basically says 
  - Please move in this direction
  - if you collide with anything, slide along it
  - make sure you're not intersecting with anything, and report everything you collide with
  - see https://www.youtube.com/watch?v=YR6Q7dUz2uk for a video explanation

## Solution

- Provide a `MoveAndSlide` `SystemParam` with an eponymous `move_and_slide` function
  - Implemented as a pure function, i.e. it does not mutate the ECS by itself
  - Returns a new position and velocity after movement and sliding along surfaces is complete
  - Every collision encountered during the slide is fed to a callback for optional processing. The callback is allowed to early-abort the slide.
  - Note: Currently, `move_and_slide` should be used to modify `Transform`, not `LinearVelocity`. The returned velocity should be stored elsewhere and passed to the next move and slide call to preserve momentum. ECS integration with `LinearVelocity` will be added later.
- High level overview: (edited by @Jondolf to match new multi-plane solver)
  1. Initial Gauss-Seidel depenetration pass
  2. For each iteration, until movement is done or max iterations reached:
     - Sweep the shape along the velocity vector
     - If we hit something, move up to the hit point
     - Collect contact planes
     - Project velocity to slide along contact planes
  3. Final Gauss-Seidel depenetration pass
- Inspirations
  - [kcc_prototypes](https://github.com/Ploruto/kcc_prototyping): general API shape, various code snippets
  - Box2D's [Character Mover](https://box2d.org/documentation/md_character.html): vector clipping and misc. ideas
  - Quake 3's [PM_SlideMove](https://github.com/id-Software/Quake-III-Arena/blob/dbe4ddb10315479fc00086f08e25d968b4b43c49/code/game/bg_slidemove.c#L45): 3D plane solver
  - [Solar Ash](https://store.steampowered.com/app/1867530/Solar_Ash/)'s implementation, as per [Tech Breakdown: Collision Sliding](https://blog.littlepolygon.com/posts/sliding/): depenetration
  - [Tweets](https://x.com/PhilippeStA/status/1233008602551590913) by Philippe St-Amand: new multi-plane solver that projects onto a convex cone, inspired by some of Philippe's ideas
- Prior work in Avian KCC tech and comparison
  - [kcc_prototypes](https://github.com/Ploruto/kcc_prototyping):
    - very similar API
    - switched to Q3 plane solver, but then a hard-drive failure erased the progress.
    - no depenetration
  - [avian_collide_and_slide](https://github.com/tracefree/avian_collide_and_slide): 
    - uses Q1 plane solving
    - no depenetration
  - [bevy_fps_controller](https://github.com/qhdwight/bevy_fps_controller): 
    - Q1/Source inspired
    - no exposed collide-and-slide
    - no plane solving
    - no depenetration
- Special thanks to @UndefinedBHVR, who spearheaded a first iteration of collide-and-slide that I was able to iterate on and who's breadcrumbs I followed a lot :heart: 
- TODO:
  - [x] Fix some jitter on plane intersections
  - [x] Write docs
  - [x] Add examples to docs
  - [x] Replace the asset links juuuust before merging
    - See https://github.com/avianphysics/avian_asset_files/pull/1

## Out of Scope for this PR

- Example KCC implementation
  - This is a deep rabbit hole and I'll leave that as a followup to not bloat this PR. 
  - See https://github.com/avianphysics/avian/issues/450 for a list of minimal character controllers that should be showcased eventually
  - For now, an integration into https://github.com/janhohenheim/character_controller_experiments will serve as proof that a KCC is doable with this API.
- Changing the existing KCC examples
  - I think these should just be entirely rewritten later on
- predicting intersections with moving bodies
- calculating forces to be exerted onto touching dynamic bodies and vice versa
- minimal examples, for now we just have the in-depth testbeds
- a more diverse 2D testbed: requires parsing SVGs like Box2D

## Testing

- [x] 3D testbed
  - `bevy run --example collide_and_slide -p avian3d`
- [x] 2D testbed
  - `bevy run --example collide_and_slide -p avian2d`
- [x] integration in https://github.com/janhohenheim/character_controller_experiments

---

## Showcase

https://github.com/user-attachments/assets/426b0e4e-73ad-490c-acc2-1489eb848066

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -369,10 +369,12 @@
 //!
 //! ## Is there a character controller?
 //!
-//! Avian does not have a built-in character controller yet, so if you need one,
-//! you will need to implement it yourself. However, third party character controllers
-//! like [`bevy_tnua`](https://github.com/idanarye/bevy-tnua) support Avian, and [`bevy_mod_wanderlust`](https://github.com/PROMETHIA-27/bevy_mod_wanderlust)
-//! and others are also likely to get Avian support soon.
+//! Avian does not have a built-in character controller yet. However, it has a [`MoveAndSlide`]
+//! system parameter with utilities for implementing your own kinematic character controllers.
+//! See its documentation for more information.
+//!
+//! There are also some third party character controllers such as [`bevy_tnua`](https://github.com/idanarye/bevy-tnua)
+//! that support Avian.
 //!
 //! For custom character controllers, you can take a look at the
 #![cfg_attr(


src/collision/collide_and_slide.rs
```diff
@@ -0,0 +1,389 @@
...
+                    if current_clip_velocity.dot(planes[j].into()) >= DOT_EPSILON {
+                        // move doesn't interact with the plane
+                        continue;
+                    }
+                    #[cfg(feature = "2d")]
+                    {
+                        // stop dead at a double plane interaction
+                        velocity = Vector::ZERO;
+                        break 'outer;
+                    }
```
unpairedbracket: Does this cause hitches on obtuse-angle seams?

src/collision/collide_and_slide.rs
```diff
@@ -0,0 +1,389 @@
...
+            if planes.len() >= config.max_planes {
+                velocity = Vector::ZERO;
+                break 'outer;
+            }
+            planes.push(Dir::new_unchecked(hit.normal1));
+
+            // modify velocity so it parallels all of the clip planes
+
+            // find a plane that it enters
+            for i in 0..planes.len() {
```
unpairedbracket: (comment applies to this entire `for in in 0..planes.len() {}` loop, but I couldn't convince github to let me select a span that long)

My (physicist's) mathematical intuition suggests this might be doable in a more robust way (hopefully one that doesn't need different implementation for 2d vs 3d) with a sprinkling of the Gram-Schmidt orthogonalisation procedure on the plane normals. It might need to be a slightly unusual version because the constraints are inequality conditions and not strict equality, but I'll see if I can figure out how to make it work if you're interested

unpairedbracket (commented):
I'm not very familiar with the prior art for this stuff so it's possible it's an already-explored dead end, but I think there might be a better way to approach the plane solver

src/collision/collide_and_slide.rs
```diff
@@ -0,0 +1,389 @@
...
+        mut on_hit: impl FnMut(CollideAndSlideHitData) -> bool,
+    ) -> CollideAndSlideResult {
+        let mut position = origin;
+        let original_velocity = velocity;
+        let mut time_left = self.time.delta_secs();
+        let mut planes = config.planes.clone();
+
+        'outer: for _ in 0..config.collide_and_slide_iterations {
+            let sweep = time_left * velocity;
+            let Some((vel_dir, speed)) = Dir::new_and_length(sweep).ok() else {
```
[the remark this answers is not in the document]
janhohenheim: https://github.com/avianphysics/avian/pull/894/commits/1eeeb7a906536edd7be4fabbc953452f7067dd67

src/collision/collide_and_slide.rs
janhohenheim: this one I will have to leave to you for sure hehe
But yes very open to alternatives!

src/character_controller/move_and_slide.rs
```diff
@@ -0,0 +1,389 @@
...
+                    if current_clip_velocity.dot(planes[j].into()) >= DOT_EPSILON {
+                        // move doesn't interact with the plane
+                        continue;
+                    }
+                    #[cfg(feature = "2d")]
+                    {
+                        // stop dead at a double plane interaction
+                        velocity = Vector::ZERO;
+                        break 'outer;
+                    }
```
unpairedbracket: Ah, I'd been imagining this as top-down 2D, running along hugging a wall. In that case I'd expect this to cause hitching whenever the wall bends inwards: stopping completely where I'd expect (as a player) to just slow down slightly, by an amount depending on how much the wall is bending.

On the other hand, I couldn't figure out how to make the idea that would have changed this work, so I can't be of much use 😅

src/character_controller/move_and_slide.rs
```diff
@@ -0,0 +1,389 @@
...
+            if planes.len() >= config.max_planes {
+                velocity = Vector::ZERO;
+                break 'outer;
+            }
+            planes.push(Dir::new_unchecked(hit.normal1));
+
+            // modify velocity so it parallels all of the clip planes
+
+            // find a plane that it enters
+            for i in 0..planes.len() {
```
unpairedbracket: Sadly I haven't quite managed to firm up this idea into anything concrete enough to implement. I'll chime in again if I figure anything out I'm sure!

unpairedbracket (approved):
Looks good to me, I think all of my (actionable) comments have been handled now!

Jondolf (approved):

PR #898, opened 2025-11-26
Scale `skin_width` by `PhysicsLengthUnit`

Jondolf:
# Objective

The `move_and_slide_2d` example sometimes gets stuck on ramps due to a skin width that is too small. This is because it's not accounting for the scale of the world.

## Solution

Scale the `skin_width` by `PhysicsLengthUnit`. This appears to fix the problem.

## Testing

Ran the `move_and_slide_2d` example.


janhohenheim (commented):
Is it intentionally not scaled here?
https://github.com/avianphysics/avian/blob/980e4c53b6b9b4c3cb2f934ec53dc1e90813d9f2/src/character_controller/move_and_slide.rs#L535

Jondolf:
> Is it intentionally not scaled here?

It's passed to `intersections`, which scales the given prediction distance. Though it is maybe a bit confusing, perhaps we shouldn't scale any method args and only scale the `skin_width` stored in the config

PR #908, opened 2025-12-10
Forces read only

### LINKED ISSUE #907: `Forces` QueryData not able to call read-only functions from a `ForcesReadOnlyItem`
When querying from a set of entities with the `Forces` QueryData, the user is unable to call any read-only functions defined on `ForcesItem`, e.g.

```rust
fn get_total_inertia(objects: Query<(Forces, &Mass)>) {
    let total = objects
        .iter()
        .map(|(forces, mass)| forces.linear_velocity() * mass.0)
        .reduce(|a, b| a + b);
}
```

returns the compilation error:
```
no method named linear_velocity found for struct forces::query_data::ForcesReadOnlyItem<'__w, '__s> in the current scope
```

A straightforward approach would be to implement each of the read-only functions from `ForcesItem` on `ForcesReadOnlyItem` [here](https://github.com/avianphysics/avian/blob/main/src/dynamics/rigid_body/forces/query_data.rs), that would return the same results as a `ForcesItem`. This would allow read-only access to `Forces` methods, freeing up time needed to mutably retrieve the components

Exotik850:
# Objective

Fixes #907

## Solution

This PR splits the `RigidBodyForces` trait into `ReadRigidbodyForces` and `WriteRigidBodyForces`, and implements `ReadRigidBodyForces` for the `ForcesReadOnlyItem` type. This allows for immutable functions to be called on a `Forces` item immutably retrieved from queries, as shown below.

## Testing

Tested by building the example shown in the issue:

```rust
fn get_total_inertia(objects: Query<(Forces, &Mass)>) {
    let total = objects
        .iter()
        .map(|(forces, mass)| forces.linear_velocity() * mass.0)
        .reduce(|a, b| a + b);
}
```
Which now compiles without errors and works as expected.

### DIFF
--- a/src/dynamics/rigid_body/forces/query_data.rs
+++ b/src/dynamics/rigid_body/forces/query_data.rs
@@ -170,7 +170,7 @@ impl ReadRigidBodyForces for ForcesReadOnlyItem<'_, '_> {}
 impl WriteRigidBodyForces for ForcesItem<'_, '_> {}
 impl WriteRigidBodyForces for NonWakingForcesItem<'_, '_> {}
 
-/// A trait for reading and writing properties of a dynamic [rigid body](RigidBody).
+/// A trait for reading and writing forces of a dynamic [rigid body](RigidBody).
 ///
 /// This is implemented as a shared interface for the [`ForcesItem`] and [`NonWakingForcesItem`]
 /// returned by [`Forces`].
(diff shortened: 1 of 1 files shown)


Jondolf (approved):
Thanks! I fixed up some of the warnings that were causing CI to fail, so this should be good to merge now :)

PR #913, opened 2025-12-26
Add joint motors for revolute and prismatic joints

### LINKED ISSUE #465: Implement joint motors

pierre-l:
# Objective

Supports velocity and position control with optional timestep-independent spring-damper parameters (frequency, damping_ratio). Includes warm starting and motor force feedback via JointForces.

**Partially** fixes #465
This implementation only works for joints with a single axis (`PrismaticJoint`, `RevoluteJoint`). 
Additional work is needed to support multiple axes (`SphericalJoint`).

## Solution

Add three `MotorModel`: `ForceBased` and `AccelerationBased` are directly inspired from the Rapier eponymous models, the `SpringDamper` model doesn't exists in Rapier but seems much easier to use to me. The latter is the default.

The motors are solved through XPBD like any other joint constraint.

Relevant Rapier sources:
- [src/dynamics/joint/motor_model.rs](https://github.com/dimforge/rapier/blob/134132900adfb73e8ae7d062b0fed83a06ab8c30/src/dynamics/joint/motor_model.rs)
- [src/dynamics/joint/generic_joint.rs](https://github.com/dimforge/rapier/blob/134132900adfb73e8ae7d062b0fed83a06ab8c30/src/dynamics/joint/generic_joint.rs)
- [dynamics/solver/joint_constraint/joint_constraint_builder.rs](https://github.com/dimforge/rapier/blob/134132900adfb73e8ae7d062b0fed83a06ab8c30/src/dynamics/solver/joint_constraint/joint_constraint_builder.rs)
- [src/dynamics/joint/revolute_joint.rs](https://github.com/dimforge/rapier/blob/134132900adfb73e8ae7d062b0fed83a06ab8c30/src/dynamics/joint/revolute_joint.rs)
- [src/dynamics/joint/prismatic_joint.rs](https://github.com/dimforge/rapier/blob/134132900adfb73e8ae7d062b0fed83a06ab8c30/src/dynamics/joint/prismatic_joint.rs)

## Breaking change

<details>
  <summary>(first commit only, outdated by the second one)</summary>
  The current implementation minimizes breaking changes, however, users who directly add the XPBD solver systems instead of using the plugin will have to make changes for motors to work:
  
  Screenshot from 2026-01-04 14-22-18
  
  Similarly, users who manually implemented `XpbdConstraint` for custom joints may need to implement `XpbdMotorConstraint` instead if they want motors to work.
</details>

Users who manually implemented `XpbdConstraint` for custom joints may need to write a custom implementation for the new `XpbdConstraint::warm_start_motors` method as its default implementation does nothing.

Users who manually add the `solve_xpbd_joint` systems instead of using the plugins may need to also add `warm_start_xpbd_motors`. (why would anyone do that though?)

## Testing

- A reasonable series of unit tests is included in the PR,
- The `joint_motors_2d` and `joint_motors_3d` examples provide an interactive way to test these,
- I used these examples to manually perform a brief test of motors+limits, no surprises there,
- I couldn't find proper test vectors so there currently is no test scenario where it is verified that a specific motor configuration reaches a specific state in a precise time interval.

---

## Showcase

<details>
  <summary>Concise showcase</summary>

```rust
// Position-controlled motor (e.g., servo arm)
RevoluteJoint::new(anchor, arm)
    .with_motor(AngularMotor {
        enabled: true,
        target_position: PI / 4.0,  // 45 degrees
        max_torque: 100.0,
        motor_model: MotorModel::SpringDamper {
            frequency: 2.0,
            damping_ratio: 1.0,  // critically damped
        },
        ..default()
    })

// Velocity or position can then be set on demand
fn set_motor_position(mut joints: Query<&mut RevoluteJoint>) {
      for mut joint in &mut joints {
          if let Some(motor) = joint.motor.as_mut() {
              motor.target_position = PI / 2.0; // 90 degrees
          }
      }
  }
```
</details>

See the examples included in this PR for more detailed examples.

[Screencast from 2026-01-11 14-55-37.webm](https://github.com/user-attachments/assets/2246dad7-2638-4741-9bf0-4d74ae5c0cee)

### DIFF
--- a/src/collision/collider/parry/mod.rs
+++ b/src/collision/collider/parry/mod.rs
@@ -1187,6 +1187,14 @@ impl Collider {
     /// The [`CollisionMargin`] component can be used to add thickness to the shape if needed.
     /// For thin shapes like triangle meshes, it can help improve collision stability and performance.
     ///
+    /// When encountering issues with
+    /// [Ghost Collisions](https://box2d.org/posts/2020/06/ghost-collisions/) where bodies collide
+    /// with the internal edges of the trimesh, try setting the flag
+    /// [`TrimeshFlags::FIX_INTERNAL_EDGES`] via [`Self::trimesh_from_mesh_with_config`], which
+    /// should fix the issue for edges where none of the adjacent triangles border their neighbor
+    /// triangles convexly (with an angle less than 180°). Note that this will also not fix the
+    /// issues with boundary edges between two different colliders.
+    ///
     /// # Example
     ///
     /// ```
@@ -1222,6 +1230,14 @@ impl Collider {
     ///
     /// The [`CollisionMargin`] component can be used to add thickness to the shape if needed.
     /// For thin shapes like triangle meshes, it can help improve collision stability and performance.
+    ///
+    /// When encountering issues with
+    /// [Ghost Collisions](https://box2d.org/posts/2020/06/ghost-collisions/) where bodies collide
+    /// with the internal edges of the trimesh, try setting the flag
+    /// [`TrimeshFlags::FIX_INTERNAL_EDGES`], which should fix the issue for edges where none of
+    /// the adjacent triangles border their neighbor triangles convexly (with an angle less than
+    /// 180°). Note that this will also not fix the issues with boundary edges between two different
+    /// colliders.
     ///
     /// # Example
     ///


pierre-l:
@Jondolf I greatly appreciate your words, thoughts, and the attention you already put into this.  

This PR's API is an aspect I'm particularly unsatisfied with yet. There is too much freedom to shoot oneself in the foot in the constructor variants, and then there's the syntax point you mentioned. I am fine with both options. I agree with the nice conceptual separation from option 2, but joint limits and motors make no sense without a joint to attach to, so it's more natural to me to have these as attributes rather than components. We agree on which is the best option then. Worst case scenario: we're both wrong on this and adjust. No big deal, the adjustment effort looks similar either way.

I should be able to bring this ready for a proper review in the next few days.

pierre-l:
I've tinkered with this for a long time and I do believe option 1 is the way to go.
It also seems to make it easier to implement joints with multiple motors like the `SphericalJoint`, however I have not been able to properly implement that yet and have very limited visibility about how much time I can spend on it.

I'm wondering if `SphericalJoint` couldn't be implemented as a composition of a `PrismaticJoint` and a `RevoluteJoint`. It seems a lot simpler to me, but maybe I'm being naive.

I updated the PR description and will mark it as ready for review.

PR #916, opened 2026-01-12
Document bevy_scene requirement for example

### LINKED ISSUE #874: Document bevy_scene requirement
Can you please document somewhere on [ColliderConstructorHierarchy](https://docs.rs/avian3d/0.4.0/avian3d/collision/collider/struct.ColliderConstructorHierarchy.html) that it will only work on a scene if the `bevy_scene` feature is enabled?

I was debugging this for many hours and stumbled on this much later than necessary.

Eterniance:
# Objective

Add the `bevy_scene` feature requirement in the [ColliderConstructorHierarchy](https://docs.rs/avian3d/0.4.1/avian3d/collision/collider/struct.ColliderConstructorHierarchy.html) documentation example. Fixes #874.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,8 @@ jobs:
                 save-if: ${{ github.ref == 'refs/heads/main' }}
                 cache-all-crates: true
                 shared-key: lints
+            - name: Install Linux dependencies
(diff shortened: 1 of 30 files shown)


Jondolf (commented):
So I think the relevant part here is that the type itself requires the `bevy_scene` feature for it to work on scenes, not that this specific example requires `bevy_scene`.

I pushed a commit to tweak the docs a bit and mention the `bevy_scene` requirement earlier on in the docs, in the section where it describes how the component interacts with `SceneRoot` :)

PR #810, opened 2025-08-22
Draft: Make spatial queries generic over colliders

NiseVoid:
# Objective

It should be possible to use the spatial query API for non-standard colliders

## Solution

- Refactor the whole thing
- Add generics everywhere

## Changelog

- The spatial query API has been reworked to allow use of non-standard collider types

## Migration Guide

TODO

PR #927, opened 2026-01-24
BVH Broad Phase and Collider Trees

Jondolf:
# Objective

Implement a new broad phase algorithm using [Bounding Volume Hierarchies (BVH)](https://en.wikipedia.org/wiki/Bounding_volume_hierarchy) provided by [OBVHS](https://github.com/DGriffin91/obvhs). This replaces our existing broad phase that uses [Sweep and Prune (SAP)](https://en.wikipedia.org/wiki/Sweep_and_prune).

The goal is to greatly reduce overhead for large scenes with lots of colliders. Updating and querying the acceleration structures should be efficient, and static or sleeping bodies should have minimal runtime cost. The existing SAP broad phase is very inefficient at this, and scales poorly even with only static geometry.

Additionally, we can reuse these BVHs for accelerating spatial queries. Currently, they just use Parry's BVH, which we update and manage very inefficiently. I will leave migrating spatial queries to the new BVHs for a follow-up however, to keep the diff more manageable.

Huge thanks to @DGriffin91 for implementing incremental leaf insertion and removal, partial rebuilds, and more for OBVHS to better suit our needs <3

## Solution

- `ColliderTree` type that contains a `Bvh2`, its collision proxies, and a workspace for reusing allocations
- `ColliderTrees` resource, containing a separate `ColliderTree` for dynamic, kinematic, static, and standalone (=no body) colliders
- `ColliderTreePlugin` that manages `ColliderTrees` for a collider type `C`
  - Adds/removes colliders to/from trees, and updates proxy data
  - Updates `ColliderAabb`s and `EnlargedAabb`s of colliders
  - Manages "moved proxies": colliders that moved past their `EnlargedAabb`, or otherwise require updating the tree
  - Optimizes trees to maintain good query performance, in an async task that runs concurrently with the narrow phase and solver
    - Full rebuilds, partial rebuilds, or reinsertion based on `ColliderTreeOptimization` settings (default is an adaptive hybrid that switches between them)
- `BroadPhaseCorePlugin` that sets up the resources, system sets, and diagnostics required for the broad phase
- `BvhBroadPhasePlugin` that implements a BVH-based broad phase algorithm
  - Traverses each tree for every moved proxy (colliders that don't move beyond their enlarged AABB don't need to check for new overlaps)
  - Creates edges in the contact graph for every new overlap between enlarged AABBs

## Testing

Ran examples and tested adding/removing/enabling/disabling colliders and changing various different configurations. Users also did some testing in real-world applications.

## Note for Reviewers

The commit history here is a bit borked for some reason, and contains some older commits that are unrelated. The first real commit is "Implement dynamic BVH broad phase with OBVHS (WIP)" (a603c5e).

---

## Showcase

Here is the new `bvh` example, for stress testing the collider trees with various different configurations.

https://github.com/user-attachments/assets/8d0e6088-c101-4e37-a597-dce356ca9e0c

### Dynamic Scene

Below is a test scene with 10k colliders with approximately 25% of them being moved each frame. Spatial queries are disabled.

Before, with the SAP broad phase:

Before

After, with the BVH broad phase:

After

Note that while the tree optimization has a fairly large cost here, in real-world applications the cost should often be more hidden, as it is done in parallel with the narrow phase and solver.

### Static Scene

Below are the results before and after, for 40k static colliders with no movement. Spatial queries are disabled. Debug rendering is also disabled, as gizmos start to tank performance at this scale.

BVH Performance

## Future Work

- Make spatial queries use the new BVH (adopt #810)
- Explore making collider trees more flexible, ex: allow users to easily create their own trees and perform queries on them
- Try doing a small amount of reinsertion on the static tree at each step to improve its query performance
- Optimize partial rebuilds further

### DIFF
--- a/crates/avian2d/Cargo.toml
+++ b/crates/avian2d/Cargo.toml
@@ -190,6 +190,10 @@ name = "revolute_joint_2d"
 required-features = ["2d", "default-collider"]
 
 [[example]]
+name = "joint_motors_2d"
+required-features = ["2d", "default-collider"]
+
+[[example]]
 name = "sensor"
 required-features = ["2d", "default-collider"]
 
--- a/crates/avian2d/examples/joint_motors_2d.rs
+++ b/crates/avian2d/examples/joint_motors_2d.rs
@@ -1,0 +1,271 @@
+//! Demonstrates motor joints in 2D.
+//!
+//! - Left side: Revolute joint with velocity-controlled angular motor (spinning wheel)
+//! - Center: Revolute joint with position-controlled angular motor (servo)
+//! - Right side: Prismatic joint with linear motor (piston)
+//!
+//! Controls:
+//! - Arrow Up/Down: Adjust left motor target velocity
+//! - A/D: Adjust center motor target angle
+//! - W/S: Adjust right motor target position
+//! - Space: Toggle motors on/off
+
+use avian2d::{math::*, prelude::*};
+use bevy::prelude::*;
+use examples_common_2d::ExampleCommonPlugin;
+
+fn main() {
+    App::new()
+        .add_plugins((
+            DefaultPlugins,
+            ExampleCommonPlugin,
+            PhysicsPlugins::default(),
+        ))
+        .insert_resource(ClearColor(Color::srgb(0.05, 0.05, 0.1)))
+        .insert_resource(Gravity(Vector::NEG_Y * 1000.0))
+        .add_systems(Startup, setup)
+        .add_systems(Update, (control_motors, update_ui))
+        .run();
+}
+
+#[derive(Component)]
+struct VelocityMotorJoint;
+
+#[derive(Component)]
+struct PositionMotorJoint;
+
+#[derive(Component)]
+struct PrismaticMotorJoint;
+
+#[derive(Component)]
+struct UiText;
+
+fn setup(mut commands: Commands) {
+    commands.spawn(Camera2d);
+
+    // === Velocity-Controlled Revolute Joint (left side) ===
+    // Static anchor for the wheel
+    let velocity_anchor = commands
+        .spawn((
+            Sprite {
+                color: Color::srgb(0.5, 0.5, 0.5),
+                custom_size: Some(Vec2::splat(20.0)),
+                ..default()
+            },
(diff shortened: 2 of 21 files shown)


eswartz:
> It's intentional that `body1` and `body2` are `None` if the colliders are not attached to a body (i.e. they are "standalone" colliders). This was apparently not the case before, but I believe that was partially an oversight, or maybe done to work around some old problem, I'm not entirely sure. Either way I would personally expect the bodies to be `None` if there is no actual `RigidBody` involved. I do need to add this to the migration guide though!

I see, that makes sense.

> Hmm 🤔 I have a test scene that is changing `RigidBody` type between dynamic, kinematic, and static for some colliding entities, but I'm not getting this panic. Any chance you could get a more minimal reproduction of this? Also make sure you have the latest version with a quick `cargo update`

I need to update the purported cause -- it apparently has nothing directly to do with swapping `RigidBody` types. I have seen the same panic a few more times over the past few days without this. IIRC it usually happens when my player mesh/KinematicBody is jumping and landing on the static RigidBody parts of the world. The similarities to the original "noclip" situation are that the player and world penetrate. 

I will work on a smaller reproducible case.

Jondolf:
Thanks, I'll try that! I did also get a repro for panics when changing `RigidBody` type, it often happens when you switch *some* dynamic bodies to kinematic or vice versa, but not all:
```text
thread 'main' (77583) panicked at /home/joona/.cargo/git/checkouts/obvhs-910b4198c23e914d/4735d72/src/bvh2/mod.rs:609:35:
index out of bounds: the len is 60 but the index is 89
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Encountered a panic in system `avian2d::collider_tree::update::update_dynamic_kinematic_aabbs<avian2d::collision::collider::parry::Collider>`!
Encountered a panic in system `avian2d::schedule::run_physics_schedule`!
Encountered a panic in system `bevy_app::main_schedule::FixedMain::run_fixed_main`!
Encountered a panic in system `bevy_time::fixed::run_fixed_main_schedule`!
Encountered a panic in system `bevy_app::main_schedule::Main::run_main`!
```
This is slightly different from the panic you got though. Either way, I'll look into these and see if I can fix them...

Edit: ah this actually looks like the same panic as your new repro

PR #933, opened 2026-01-31
Return `MoveAndSlideHitReponse` enum from `on_hit` callback

Jondolf:
# Objective

The callbacks for `move_and_slide` and `intersections` don't currently work like their docs describe. The former should abort when the callback returns false (currently it just ignores a single hit), and the latter doesn't actually do anything with the returned boolean.

For `move_and_slide`, the callback should ideally return an enum with three variants: accept the hit, ignore the hit but continue, and ignore the hit and abort. For `intersections`, the callback should still return a boolean, but it should actually abort when `false` is returned.

## Solution

Add a `MoveAndSlideHitResponse` enum:

```rust
/// Indicates how to handle a hit detected during [`MoveAndSlide::move_and_slide`].
///
/// This is returned by the `on_hit` callback provided to [`MoveAndSlide::move_and_slide`].
#[derive(Debug, PartialEq)]
pub enum MoveAndSlideHitResponse {
    /// Accept the hit and continue the move and slide algorithm.
    Accept,

    /// Ignore the hit and continue the move and slide algorithm.
    ///
    /// Note that the shape will still be moved up to the point of collision,
    /// but the velocity will not be modified to slide along the surface.
    Ignore,

    /// Ignore the hit and abort the move and slide algorithm.
    ///
    /// Note that the shape will still be moved up to the point of collision,
    /// but no further movement or velocity modification will be performed.
    Abort,
}
```

This is handled as the docs describe.

Also make `intersections` abort further hit processing when the callback returns `false`.


src/character_controller/move_and_slide.rs
```diff
@@ -530,19 +551,24 @@ impl<'w, 's> MoveAndSlide<'w, 's> {
...
                 position: &mut position,
                 velocity: &mut velocity,
-            }) {
+            });
+
+            if hit_response == MoveAndSlideHitResponse::Accept {
                 planes.push(Dir::new_unchecked(sweep_hit.normal1.f32()));
+            } else if hit_response == MoveAndSlideHitResponse::Abort {
+                break;
             }
```
janhohenheim: how come an ignore doesn't `continue` here?

Jondolf: `Ignore` skips this specific surface (i.e. doesn't push the plane like `Accept` does) but continues processing other hits. The rest of the hits here are processed below in `intersections`.

The semantics here are a bit confusing since `Ignore` doesn't ignore the actual shape cast hit. The shape is still moved up to the surface, but the return value of the callback just determines what to do with each plane found at that point, which affects the velocity projection. Having it ignore the shape cast hit would require a filter on the shape cast itself (tbf we *should* have this too), not on the collection of collision planes.

Tangentially related: I wonder if using `Abort` on the first hit is actually equivalent to [`move_and_collide`](https://docs.godotengine.org/en/stable/classes/class_physicsbody3d.html#class-physicsbody3d-method-move-and-collide) in Godot, or [`slide: false`](https://docs.rs/rapier3d/latest/rapier3d/control/struct.KinematicCharacterController.html#structfield.slide) in Rapier :thinking: I think it should be; it stops at the first hit and then aborts move-and-slide without any sliding

janhohenheim (approved):

PR #962, opened 2026-03-20
2D picking with a 3D camera

rdrpenguin04:
# Objective

Allow using perspective cameras with the 2D physics plugin

## Solution

Follow the ray that `bevy_picking` gives to its intersection with the Z axis rather than truncating the value to 2D

## Testing

(see below)

---

## Showcase

I added a demo to the repository under the name `picking_2d_in_perspective.rs` (in `avian2d`); it looks a bit like this:

[Screencast_20260320_083503.webm](https://github.com/user-attachments/assets/a2db47e8-0b60-47c7-9be7-05984c14981b)

I had hoped to add a free camera to the demo so you could see rotations working as well; as of yet, I haven't gotten around to it. I may finish that in a separate PR if I have the time, or someone else can add it in :smile:

### DIFF
--- a/crates/avian2d/Cargo.toml
+++ b/crates/avian2d/Cargo.toml
@@ -173,6 +173,10 @@ name = "one_way_platform_2d"
 required-features = ["2d", "default-collider"]
 
 [[example]]
+name = "picking_2d_in_perspective"
+required-features = ["2d", "debug-plugin", "bevy_picking"]
+
+[[example]]
 name = "prismatic_joint_2d"
 required-features = ["2d", "default-collider"]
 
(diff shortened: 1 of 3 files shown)


Jondolf (approved):
Thanks!

Luracasmus:
Does this mean that colliders must be located at Z = 0 (world space?) to be picked properly with a camera using perspective projection, and if so, would it be possible to allow that Z level to be configured instead?

Luracasmus:
> Does this mean that colliders must be located at Z = 0 (world space?) to be picked properly with a camera using perspective projection, and if so, would it be possible to allow that Z level to be configured instead?

I made something that seems to work in my use case (3D camera with perspective projection and colliders located far away in negative Z, which was broken earlier) but I don't know if it's fully correct
https://github.com/Luracasmus/avian/commit/4c2e1e1d3a64418bf6d3af27fe2c5c2c2a69f9d0

PR #964, opened 2026-03-25
perf: draw HeightField debug as grid wireframe instead of per-triangl…

PandaStacker:
The 3D HeightField debug rendering previously created a SharedShape::new(triangle) for every triangle and recursed into draw_collider. For a 32×32 heightfield this means ~1922 heap allocations per collider per frame, making debug rendering unusable for chunked terrain with multiple heightfield colliders.

This replaces the per-triangle approach with direct grid wireframe drawing using the heightfield's heights matrix, scale, and dimensions. The result is visually identical but avoids all per-triangle allocations.

Tested on a project with a basic HeightField mesh that was tanking to around 20 FPS. Now hitting the cap at 120 FPS.

### DIFF
--- a/src/debug_render/gizmos.rs
+++ b/src/debug_render/gizmos.rs
@@ -270,13 +270,45 @@ impl PhysicsGizmoExt for Gizmos<'_, '_, PhysicsGizmos> {
                     );
                 }
                 #[cfg(feature = "3d")]
-                for triangle in s.triangles() {
-                    self.draw_collider(
-                        &Collider::from(SharedShape::new(triangle)),
-                        position,
-                        rotation,
-                        color,
-                    );
+                {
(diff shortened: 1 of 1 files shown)


Jondolf (commented):
Hey! Like Bevy, we don't accept Claude commits (see Bevy's [AI policy](https://bevy.org/learn/contribute/policies/ai/)). We should have this stated somewhere explicitly though, that's my bad.

Additionally, this doesn't actually work correctly for all height fields. This assumes that every cell is a rectangle, but `HeightField` supports removing cells or even individual triangles by configuring cell statuses. The original implementation with the `triangles` iterator should handle this properly, but just treating the height field as a grid does not.

The performance of this could definitely be improved though, creating a new SharedShape for each triangle is silly

PR #965, opened 2026-03-27
chore: migrate to Bevy 0.19-dev

jbuehler23:
Migrate avian3d and its subcrate deps to Bevy 0.19-dev.

### DIFF
--- a/src/collision/collider/collider_hierarchy/mod.rs
+++ b/src/collision/collider/collider_hierarchy/mod.rs
@@ -45,7 +45,7 @@ use bevy::{
 ///
 /// [`Relationship`]: bevy::ecs::relationship::Relationship
 #[derive(Component, Clone, Copy, Debug, PartialEq, Eq, Reflect)]
-#[component(immutable, on_insert = <ColliderOf as Relationship>::on_insert, on_replace = <ColliderOf as Relationship>::on_replace)]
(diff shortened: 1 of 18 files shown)


src/collision/collider/collider_hierarchy/mod.rs
```diff
@@ -69,6 +69,8 @@ impl FromWorld for ColliderOf {
 impl Relationship for ColliderOf {
     type RelationshipTarget = RigidBodyColliders;
 
+    const ALLOW_SELF_REFERENTIAL: bool = true;
```
janhohenheim, line 72: this entire custom implementation can now be removed because we have upstream self-relations

src/math/mod.rs
```diff
@@ -96,6 +96,87 @@ pub(crate) type Isometry = Isometry2d;
 #[cfg(feature = "3d")]
 pub(crate) type Isometry = Isometry3d;
 
+// Conversion utilities between parry's glam types (glam 0.30 via glamx)
+// and bevy's glam types (glam 0.32). These have identical memory layouts
+// but are different types due to the version mismatch.
+#[cfg(any(feature = "parry-f32", feature = "parry-f64"))]
+pub(crate) mod parry_conv {
```
janhohenheim, line 103: this would probably look nicer if it was an extension trait

Jondolf: or we just need to wait for parry to upgrade (or PR it for them) so it uses the same glam version, which is needed for bevy_rapier to update too

Jondolf:
Closing in favor of #991

PR #966, opened 2026-03-27
allow accessing entity from intersections callback

PROMETHIA-27:
# Objective

Adds a new parameter to the callback in `MoveAndSlide::intersections`, to allow accessing the intersected entity.

## Solution

Adds a new parameter to the relevant callback.

## Testing

The change is extremely simple so as long as the logic is correct, it shouldn't need extensive testing.
I ran `cargo test` to try and catch all places that needed to be updated in the repo, and use the same logic in a vendored version of the function in a project and it has worked correctly so far.

I did see 4 failures in the `cargo test` but they seemed to be minor precision issues; float equality comparisons off by 0.0001 or so. Probably unrelated to the changes in this PR.


Jondolf (approved):
Thanks, LGTM once CI passes. Looks like a trivial error with a doc test missing the parameter

PR #970, opened 2026-04-11
Make `SpatialQueryPlugin` configurable

### LINKED ISSUE #969: Ray casting breaks breaks when time dilation is set to 0
In my game, I support pausing. When the game is paused, time dilation is set to 0, so everything freezes. `FixedUpdate` doesn't run.

However, users can still move objects during pause.

To pick objects, I use Avian, and I'd like to keep it functional during pause. Just for testing, I set the Avian schedule to `PostUpdate`, but it still breaks. In the video, you can see that I can move objects freely while the game is unpaused. But once I pause the game, I can move any object only once. After that, hovering over the object does nothing (you can normally see a white outline when the object is hovered).

https://github.com/user-attachments/assets/a3b68134-e81f-41ad-95b4-217f2166213e

The behavior on the be fixed by adding `update_moved_collider_aabbs` to `PostUpdate` manually.

However, I noticed that `RayCast` is still broken because its global direction is not update. The video above works only because it uses picking.

It would be nice to properly support use cases like this.

Shatur:
# Objective

- Fixes #969.

## Solution

I did what was suggested in #969.
The plugin now accepts a configurable schedule and `SpatialQuerySystems` now its own dedicated label.

The default behavior didn't change.
But now I can disable `SpatialQueryPlugin` from the `PhysicsPlugins` and add it manually with something like `PostUpdate`.
This fixes my problem.

## Testing

I tested it in my game.

### DIFF
--- a/migration-guides/0.6-to-main.md
+++ b/migration-guides/0.6-to-main.md
@@ -6,6 +6,7 @@ since the latest release. These guides are evolving and may not be polished yet.
 See [migration-guides/README.md](./README.md) and existing entries for information about Avian's
 migration guide process and what to put here.
 
-## `SpatialQueryPlugin` schedule now can be configured
+## `SpatialQueryPlugin` schedule is configurable
 
-Similar to other plugins, it now stores a schedule label and affected by the label passed into `PhysicsPlugin::new`.
+Similar to many other plugins, the `SpatialQueryPlugin` now stores a schedule label,
+which by default is also passed via `PhysicsPlugins::new`.
--- a/migration-guides/0.6-to-main.md
+++ b/migration-guides/0.6-to-main.md
@@ -5,3 +5,7 @@ since the latest release. These guides are evolving and may not be polished yet.
 
 See [migration-guides/README.md](./README.md) and existing entries for information about Avian's
 migration guide process and what to put here.
+
+## `SpatialQueryPlugin` schedule now can be configured
+
+Similar to other plugins, it now stores a schedule label and affected by the label passed into `PhysicsPlugin::new`.
--- a/src/schedule/mod.rs
+++ b/src/schedule/mod.rs
@@ -100,7 +100,6 @@ impl Plugin for PhysicsSchedulePlugin {
                     PhysicsStepSystems::NarrowPhase,
                     PhysicsStepSystems::Solver,
                     PhysicsStepSystems::Sleeping,
-                    PhysicsStepSystems::SpatialQuery,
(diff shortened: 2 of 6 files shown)


Shatur:
@Jondolf is there anything left I need to do in order to get this merged?

migration-guides/0.6-to-main.md
```diff
@@ -5,3 +5,7 @@ since the latest release. These guides are evolving and may not be polished yet.
 
 See [migration-guides/README.md](./README.md) and existing entries for information about Avian's
 migration guide process and what to put here.
+
+## `SpatialQueryPlugin` schedule now can be configured
+
+Similar to other plugins, it now stores a schedule label and affected by the label passed into `PhysicsPlugin::new`.
```
Jondolf:
```suggestion
## `SpatialQueryPlugin` schedule is configurable

Similar to many other plugins, the `SpatialQueryPlugin` now stores a schedule label,
which by default is also passed via `PhysicsPlugins::new`.
```

src/schedule/mod.rs
Jondolf:
```suggestion
/// 6. Finalize
```

src/spatial_query/mod.rs
```diff
@@ -200,6 +223,12 @@ impl Plugin for SpatialQueryPlugin {
     }
 }
 
+/// Responsible for spatial queries like [raycasting](`RayCaster`) and shapecasting.
```
Jondolf:
```suggestion
/// Responsible for updating spatial query components like [`RayCaster`] and [`ShapeCaster`].
```

src/spatial_query/mod.rs
```diff
@@ -168,19 +168,42 @@ pub use shape_caster::*;
...
-        let physics_schedule = app
-            .get_schedule_mut(PhysicsSchedule)
-            .expect("add PhysicsSchedule first");
-
-        physics_schedule.add_systems(
+        app.configure_sets(
+            self.schedule,
+            SpatialQuerySystems.after(TransformSystems::Propagate),
+        )
+        .add_systems(
```
Jondolf: Just a stylistic thing, I like to separate system set config from adding systems
```suggestion
        );

        app.add_systems(
```

Jondolf (commented):
I made a few doc tweaks locally, but can't seem to push to this branch 🤔 welp, I just added my changes as review suggestions :)

Other than that, LGTM!

Jondolf (approved):

PR #971, opened 2026-04-11
Add `RayCaster::get(_global)_point`

### LINKED ISSUE #968: Add `get_point` helper to `RayCaster`
Right now to get position users need to take origin, add direction and multiply by distance:
```rust
fn print_hits(query: Query<(&RayCaster, &RayHits)>) {
    for (ray, hits) in &query {
        for hit in hits.iter() {
            println!(
                "Hit entity {} at {}",
                hit.entity,
                ray.origin + *ray.direction * hit.distance,
            );
        }
    }
}
```

But it would be nice to have a helper like [`Ray3d::get_point`](https://docs.rs/bevy/latest/bevy/prelude/struct.Ray2d.html#method.get_point).

Shatur:
# Objective

Fixes #968.

## Solution

Add the mentioned methods.

## Testing

I didn't test it, but the change is trivial. I did similar to what Bevy does:

https://docs.rs/bevy_math/0.18.1/src/bevy_math/ray.rs.html#97

### DIFF
--- a/src/spatial_query/ray_caster.rs
+++ b/src/spatial_query/ray_caster.rs
@@ -283,13 +283,13 @@ impl RayCaster {
     /// Returns the point at a given distance along the ray.
     #[must_use]
     pub fn get_point(&self, distance: f32) -> Vector {
-        self.origin + self.direction * distance
+        self.origin + *self.direction * distance
     }
 
     /// Like [`Self::get_point`], but returns the point in global coordinates.
     #[must_use]
     pub fn get_global_point(&self, distance: f32) -> Vector {
-        self.global_origin + self.global_direction * distance
+        self.global_origin + *self.global_direction * distance
     }
 }
 
--- a/src/spatial_query/ray_caster.rs
+++ b/src/spatial_query/ray_caster.rs
@@ -279,6 +279,18 @@ impl RayCaster {
             ));
         }
     }
+
+    /// Returns the point at a given distance along the ray.
+    #[must_use]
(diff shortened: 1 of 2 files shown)


src/spatial_query/ray_caster.rs
```diff
@@ -279,6 +279,18 @@ impl RayCaster {
...
+    #[must_use]
+    pub fn get_point(&self, distance: f32) -> <Vector as AsF32>::F32 {
+        self.origin.f32() + self.direction * distance
+    }
+
+    /// Like [`Self::get_point`], but returns the point in global coordinates.
+    #[must_use]
+    pub fn get_global_point(&self, distance: f32) -> <Vector as AsF32>::F32 {
+        self.global_origin.f32() + self.global_direction * distance
+    }
```
Jondolf, line 293: Given that we currently use the f32/f64 feature-dependent types elsewhere, I think this should use them too. But at some point we should definitely reconsider what features supports double precision and what features always uses f32
```suggestion
    /// Returns the point at a given distance along the ray.
    #[must_use]
    pub fn get_point(&self, distance: Scalar) -> Vector {
        self.origin + self.direction.adjust_precision() * distance
    }

    /// Like [`Self::get_point`], but returns the point in global coordinates.
    #[must_use]
    pub fn get_global_point(&self, distance: Scalar) -> Vector {
        self.global_origin + self.global_direction.adjust_precision() * distance
    }
```

Jondolf (approved):
Thanks!

PR #975, opened 2026-04-18
Make SpatialQuery more generic across colliders

NiseVoid:
# Objective

`SpatialQuery` cannot be used with non-parry collider types.

## Solution

Add a trait to let types use `SpatialQuery`, and add a generic param in a few places

## Testing

- Run various examples that use `SpatialQuery` to ensure nothing broke
- Make sure things still compile without default features

## Next steps

- Make `ShapeCast` and `RayCast` components work with the new approach
- Look into simplify some of the methods by applying translation/rotation/scaling externally
- Add an example for a custom `QueryCollider` impl
- Look into allowing `QueryCollider` without `AnyCollider`
- Use some better Rust magic to clean up the trait madness in this PR

### DIFF
--- a/src/collision/collider/mod.rs
+++ b/src/collision/collider/mod.rs
@@ -55,7 +55,7 @@ pub trait IntoCollider<C: AnyCollider> {
 
 /// Context necessary for operations involving an [`AnyCollider`].
 #[derive(Deref)]
-pub struct ColliderContext<'a, 'w, 's, T: ReadOnlySystemParam, E: EntityUsage> {
+pub struct ColliderContext<'a, 'w, 's, T: ReadOnlySystemParam, E: EntityUsage = ()> {
     /// The collider entity involved in the operation.
     pub collider: E,
     #[deref]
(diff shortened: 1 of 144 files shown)


src/spatial_query/system_param.rs
```diff
@@ -57,13 +63,14 @@ use parry::query::ShapeCastOptions;
 /// }
 /// ```
 #[derive(SystemParam)]
-pub struct SpatialQuery<'w, 's> {
-    colliders: Query<'w, 's, (&'static Position, &'static Rotation, &'static Collider)>,
+pub struct SpatialQuery<'w, 's, C: QueryCollider> {
```
Jondolf, line 66: Can we make this default to the built-in `Collider` when the feature is enabled? Alternate collider backends feel like an advanced use case and not something that should bleed into the common API, if possible

src/collision/collider/mod.rs
```diff
@@ -123,9 +136,33 @@ impl ContactManifoldContext<'_, '_, '_, ()> {
...
+        e
+    }
+}
+
+/// A type indicating no entity is used for a [`BoundedCollider`]'s `Context`
+pub struct NoEntity;
+impl EntityUsage for NoEntity {
+    type Entity = ();
+    fn to_field(_: Entity) {}
+}
```
Jondolf, line 161: This feels quite nasty, but I can see why something like this may be needed for `AabbContext` 🤔 I'll have to think if there's a better way to handle this

Jondolf (commented):
Thanks! This largely looks like the right shape for this sort of thing. I did my first pass on this, initial thoughts are mainly that some of the APIs and trait bounds may need some more thought to figure out if we can make them a bit nicer somehow

PR #977, opened 2026-04-29
Granular configuration of debug render systems

barsoosayque:
# Objective

- Enable/disable debug render systems via `run_if` conditions.
- Configure what parts of the physics should be rendered, e.g. only colliders and body axes.
- Can be configured both in runtime and before the app start.

## Solution

New resource `PhysicsDebugRenderConfig` with a bunch of bools for each group of render debug systems. By default, everything is enabled.

Alternative solution is to cram everything into `PhysicsGizmos`, but I think it's big enough already ?

## Testing

I tested this with `cargo run --example move_and_slide_3d --features use-debug-plugin` and toggling colliders rendering:

```diff
index 113216a..f4db8db 100644
--- a/crates/examples_common_3d/src/lib.rs
+++ b/crates/examples_common_3d/src/lib.rs
@@ -44,6 +44,16 @@ impl Plugin for ExampleCommonPlugin {
         if !app.is_plugin_added::<PhysicsDebugPlugin>() {
             app.add_plugins(PhysicsDebugPlugin);
         }
+
+        app.add_systems(
+            Update,
+            (
+                (|mut r: ResMut<PhysicsDebugRenderConfig>| {
+                    r.enable_colliders = !r.enable_colliders
+                })
+                    .run_if(input_just_pressed(KeyCode::KeyC)),
+            ),
+        );
     }
 }
```

### DIFF
--- a/src/debug_render/configuration.rs
+++ b/src/debug_render/configuration.rs
@@ -15,7 +15,7 @@ use bevy::{color::palettes::css::*, prelude::*};
 /// #[cfg_attr(feature = "3d", doc = "use avian3d::prelude::*;")]
 /// use bevy::prelude::*;
 ///
-/// fn disable_aabb(mut config: ResMut<GlobalPhysicsDebugConfig>) {
+/// fn disable_aabb(mut config: ResMut<PhysicsDebugRenderConfig>) {
 ///     config.enable_aabb = false;
 /// }
 /// ```
--- a/src/debug_render/configuration.rs
+++ b/src/debug_render/configuration.rs
@@ -11,8 +11,8 @@ use bevy::{color::palettes::css::*, prelude::*};
 /// For example, to disable AABB rendering:
 ///
 /// ```no_run
-/// #[cfg_attr(feature = "2d", doc = "use avian2d::prelude::*;")]
-/// #[cfg_attr(feature = "3d", doc = "use avian3d::prelude::*;")]
+#[cfg_attr(feature = "2d", doc = "use avian2d::prelude::*;")]
+#[cfg_attr(feature = "3d", doc = "use avian3d::prelude::*;")]
 /// use bevy::prelude::*;
 ///
 /// fn disable_aabb(mut config: ResMut<PhysicsDebugRenderConfig>) {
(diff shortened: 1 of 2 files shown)


Jondolf (commented):
Hmm so currently you can achieve pretty much the same thing by just configuring `PhysicsGizmos`:

```rust
app.add_systems(
    Update,
    ((|mut config_store: ResMut<GizmoConfigStore>| {
        let (_, gizmos) = config_store.config_mut::<PhysicsGizmos>();
        if gizmos.collider_color.is_none() {
            gizmos.collider_color = Some(ORANGE.into());
        } else {
            gizmos.collider_color = None;
        }
    })
    .run_if(input_just_pressed(KeyCode::KeyC)),),
);
```

I think the main problem there is that when you disable the gizmo by setting the color to `None`, you lose the old value. Ideally, the color and whether the gizmo is enabled should be distinct properties, instead of just using `Option<Color>`.

This PR does fix that problem, but I don't like splitting configuration into two separate places like this. It also means that there are now two different ways of disabling gizmos.

I think a reasonable alternative could be something like adding a `DebugColor` struct that stores a `Color` and whether it is actually rendered:

```rust
pub struct DebugColor {
    pub color: Color,
    pub is_enabled: bool,
}
```

with some constructors and easy conversion using `From<Color>`. Alternatively, we could rely on the `alpha` of the color, but that feels less obvious for users, and also makes semi-transparent colors not retain their alpha when disabling that gizmo.

barsoosayque:
> This PR does fix that problem, but I don't like splitting configuration into two separate places like this. It also means that there are now two different ways of disabling gizmos.

Fair enough, I don't like this too. I'm not super sure about gizmo configuration, can it be disabled per group like in this PR ? If so, I can migrate it to use gizmo config.

The main thing I need is not disabling gizmos, but systems that render those gizmos (well, queue gizmos to render). This is because in our project, I use debug rendering on a very minimal subset of avian, which doesn't run any simulation, but only renders colliders and axes (i.e. on the client binary, full avian simulation is on the server). Other debug rendering systems other than colliders/axes use a bunch of resources that I have to init so the project doesn't crash.

Jondolf:
You could add run conditions for the systems yes, accessing `PhysicsGizmos` in the same way as in my code example

PR #981, opened 2026-05-05
Do not force all meshes visible on gizmo settings change.

eswartz:
# Objective

- I have a scene with certain Mesh instances inside containers marked Visibility::Hidden, and a (now-fixed) issue where I was triggering a change to the `PhysicsGizmos` `GizmoConfigStore`. This was causing `change_mesh_visibility` to be called and modifying all the meshes to be visible.

## Solution

- Switch to using `Visibility::Inherited` to respect parent (in)visibility, as `Visibility::Visible` affects everything, even in hidden hierarchies.
- Since this code seems quite generic, I suppose the intent of may have been to intentionally expose all meshes for some reason, but that seems unlikely (debug rendering and the gizmos for colliders is "enough" IMO).

## Testing

- Tested with my code that toggles `PhysicsGizmos` enablement. It no longer reveals shows invisible meshes.

### DIFF
--- a/migration-guides/0.6-to-main.md
+++ b/migration-guides/0.6-to-main.md
@@ -16,3 +16,14 @@ simply ignore the new parameter with `_`.
 
 Similar to many other plugins, the `SpatialQueryPlugin` now stores a schedule label,
 which by default is also passed via `PhysicsPlugins::new`.
+
+## Physics Debug Gizmo Mesh Visibility
+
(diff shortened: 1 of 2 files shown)


Jondolf (approved):
This isn't a perfect solution, as it still doesn't respect the old value before the visibility was overridden for debug purposes, but I do think this is a better default. In the future, I'd like to explore replacing this with a dedicated "debug view" based on `RenderLayers`, which would be less destructive.

I added a migration guide, since this is considered a breaking change

PR #982, opened 2026-05-06
Make 2D picking Z plane configurable

Luracasmus:
# Objective

- Even after #962, 2D picking only works reliably when pickable colliders are located on the Z = 0 plane when e.g. using perspective projection or a rotated camera. Allowing this Z coordinate to be configured could be useful.

## Solution

- Make the Z level used for picking ray intersection configurable by adding a new field to `PhysicsPickingSettings`.
- Extend the `picking_2d_in_perspective` example with controls for moving the physics picking Z plane, rotating the camera, and switching projection type, demonstrating this new functionality.
- Write a migration guide entry for the changes in this PR and #962.

## Testing

- The extended `picking_2d_in_perspective` example seems to work properly.
- The solution works in my use case, picking from a 3D camera with perspective projection and colliders located far away in negative Z, which was previously broken.
- `cargo test` returns 4 failures, but I get the same result on the latest commit before my changes.
  ```log
  ---- dynamics::rigid_body::mass_properties::tests::mass_properties_add_remove_collider stdout ----
  
  thread 'dynamics::rigid_body::mass_properties::tests::mass_properties_add_remove_collider' (22023) panicked at crates/avian2d/../../src/dynamics/rigid_body/mass_properties/mod.rs:736:9:
  assertion `left == right` failed
    left: 15.707963
   right: 15.707964
  
  ---- dynamics::rigid_body::mass_properties::tests::mass_properties_no_auto_mass_add_remove stdout ----
  
  thread 'dynamics::rigid_body::mass_properties::tests::mass_properties_no_auto_mass_add_remove' (22026) panicked at crates/avian2d/../../src/dynamics/rigid_body/mass_properties/mod.rs:890:9:
  assertion `left == right` failed
    left: 14.999999
   right: 15.0
  
  ---- dynamics::rigid_body::mass_properties::tests::mass_properties_move_child_collider stdout ----
  
  thread 'dynamics::rigid_body::mass_properties::tests::mass_properties_move_child_collider' (22025) panicked at crates/avian2d/../../src/dynamics/rigid_body/mass_properties/mod.rs:843:9:
  assertion `left == right` failed
    left: 14.999999
   right: 15.0
  
  ---- dynamics::rigid_body::mass_properties::tests::mass_properties_rb_collider_with_set_mass_and_child_collider_with_set_mass stdout ----
  
  thread 'dynamics::rigid_body::mass_properties::tests::mass_properties_rb_collider_with_set_mass_and_child_collider_with_set_mass' (22033) panicked at crates/avian2d/../../src/dynamics/rigid_body/mass_properties/mod.rs:629:9:
  assertion `left == right` failed
    left: 14.999999
   right: 15.0
  ```

All testing done was on x86_64-unknown-linux-gnu (latest CachyOS Desktop as of writing, using the [rust](https://packages.cachyos.org/package/cachyos-extra-znver4/x86_64_v4/rust) package from the `cachyos-extra-znver4` distribution repository).

### DIFF
--- a/crates/avian2d/examples/picking_2d_in_perspective.rs
+++ b/crates/avian2d/examples/picking_2d_in_perspective.rs
@@ -33,8 +33,11 @@ fn setup(
         Transform::from_xyz(0.0, 0.0, 1.0), // We have to start slightly away from the Z axis, since now everything is drawing at Z = 0
     ));
 
+    // For example purposes, define a custom picking plane that can be moved with the E and Q keys.
+    // Using a picking plane like this is not required, but the Z coordinates of 2D colliders
+    // should match `PhysicsPickingSettings::z_plane` for them to be picked at the correct XY coordinates.
     commands
-        .spawn((PickingPlane, Transform::default(), Visibility::Visible)) // [`Transform::default()`] is at Z = 0, which matches the default `z_plane` in [`PhysicsPickingSettings`]
+        .spawn((PickingPlane, Visibility::Visible, Transform::default()))
         .with_children(|child_spawner_commands| {
             let square_sprite = Sprite {
                 color: Color::srgb(0.7, 0.7, 0.8),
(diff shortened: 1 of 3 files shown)


Jondolf (approved):
Thanks! This seems like a reasonable addition to me. A bit of a shame that the shapes must be at the same z coordinate, but fixing that properly seems a bit tricky 🤔 so this is a fine compromise for now.

I tweaked the docs and migration guide a bit, just to improve clarity and wording at some parts

PR #987, opened 2026-05-28
Bevy 0.19 work in progress: bump parry to 0.27 and sync with latest bevy main

stuartparmenter:
Work to get avian compiling on bevy 0.19 (bevy main).

Builds on top of #965 (the in-progress bevy 0.19 migration). The first three commits here are the same as #965. On top of those, this PR adds:

* Bump parry to 0.27. This bumps parry's transitive glamx to 0.2 (glam 0.32), matching bevy main. The `parry<->bevy` math conversion helpers (`vec*_from_parry`, `vec*_to_parry`, `quat_*_parry`) and the `math::parry_conv` module can all be removed.
* Sync avian to the latest bevy main: fixes breakage from bevy PR #24164 and the `bevy_scene` to `bevy_world_serialization` feature rename.

avian2d also builds again as a side effect, since most of its remaining errors were 2D code calling `vec3_*` conversion helpers that only existed under `#[cfg(feature = "3d")]`.

Opening as draft since #965 hasn't landed yet and this should probably go on top of it.

(Note: LLM assisted - Opus)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -435,7 +435,7 @@ dependencies = [
 name = "avian_derive"
 version = "0.2.2"
 dependencies = [
- "proc-macro-error2",
+ "proc-macro-error3",
  "proc-macro2",
  "quote",
  "syn",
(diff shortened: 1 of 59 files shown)


Jondolf:
Closing in favor of #991, which is more up-to-date and cleaned up and has a migration guide etc. I've added you as a co-author on the merge commit though, since it was building on top of this PR

PR #991, opened 2026-06-03
Update to Bevy 0.19.0-rc

olukowski:
Based on the code changes in #987, but pinned to the 0.19 rc instead of main.

Also fixes CI issues.

### DIFF
--- a/benches/Cargo.toml
+++ b/benches/Cargo.toml
@@ -31,13 +31,13 @@ avian3d = { path = "../crates/avian3d", default-features = false, features = [
 ], optional = true }
 
 # Bevy
-bevy = { version = "0.18.0", default-features = false }
+bevy = { version = "0.19.0-rc", default-features = false }
 
 # CLI
-clap = { version = "4.5", features = ["derive"] }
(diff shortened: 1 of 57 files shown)


Jondolf (approved):
Thanks! As you might've noticed, I pushed quite a few commits to fix and clean up some stuff, get CI passing, update dependencies, and add a migration guide :)

We still need to update to the full 0.19 release, and actually publish a new Avian release, but I'll do that in a follow-up

benches/Cargo.toml
```diff
@@ -31,13 +31,13 @@ avian3d = { path = "../crates/avian3d", default-features = false, features = [
 ], optional = true }
 
 # Bevy
-bevy = { version = "0.18.0", default-features = false }
+bevy = { version = "0.19.0-rc", default-features = false }
```
martinfrances107, line 34: This can now be 0.19.0 not 0.19.0-rc as the release has happened

Jondolf: Yeah I'll change it to 0.19.0 in a follow-up PR where I also update the git deps to released versions (I need to actually release new versions of those other crates first)

PR #999, opened 2026-06-15
fix(debug): concrete types instead of SharedShape

Jengamon:
# Objective

For debug gizmos, as mentioned in https://github.com/avianphysics/avian/pull/964#pullrequestreview-4289833748, it is a little silly to allocate an Arc per shape (what SharedShape is a wrapper around).

## Solution

Copy the shape data into a bevy primitive and use the From impl for primitives that implement "IntoCollider" to create the colliders instead.

Also add the missing Triangle3d IntoCollider impl

## Testing

- Did you test these changes? If so, how?
No, not yet.
- Are there any parts that need more testing?
It should be a relatively straightforward change, so no afaict.
- How can other people (reviewers) test your changes? Is there anything specific they need to know?
Simply point to the branch and use the PhysicsDebugPlugin, there should be no change in behavior if successful.
- If relevant, what platforms did you test these changes on, and are there any important ones you can't test?
Not relevant


Jengamon:
oh, also instead of just replacing the usage for heightfields, replace *all* uses of SharedShape, b/c debugging shouldn't need to allocate an Arc for each element to be debugged.

Also, I haven't found an MSRV policy for avian afaict, so I'm following bevy's MSRV policy, which is in essence "latest stable", which is why I use `cfg_select!`

Jengamon:
marking as draft b/c I did a silly and forgot that bevy primitives only come in single-precision (Vec{2,3})...

will check out writing a PR for adding double-precision primitives (DVec{2,3})

EDIT: I am seely, and ~~adjust_precision~~AsF32::f32 exists for a reason

PR #1019, opened 2026-07-03
Add VariantDefaults to RigidBody

larsraph:
# Objective

`bsn!{}` doesn't work with `RigidBody` because enums components that don't have default values for each variant. 

Behind the hood this seems like running into a bevy limitation thats a workaround for a rust limitation. 

Other enum components may have this same issue but I didn't check. 

## Solution

Derive `VariantDefaults` on `RigidBody`.

### DIFF
--- a/src/dynamics/rigid_body/mod.rs
+++ b/src/dynamics/rigid_body/mod.rs
@@ -24,7 +24,7 @@ use crate::{
     prelude::{forces::AccumulatedLocalAcceleration, *},
 };
 use bevy::{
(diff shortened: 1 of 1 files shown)


Exotik850:
Working with the new scenes and avian in a project atm, this would be a great help

DavidCrossman:
Current workaround is `template_value(RigidBody::Dynamic)`.
`VariantDefaults` is removed in Bevy 0.20 (which now has a release candidate) and enum variants no longer need to have default values, so this PR can probably be closed.

PR #1023, opened 2026-07-07
Fix fast-spinning balls sinking into the ground

### LINKED ISSUE #1056: Fast-spinning balls sinking into the ground
**Issue:**

a 16 mm marble at 1 m/s rides 55% of its radius deep at 60 Hz.

The substep separation update rotates contact anchors as material points, but a ball's contact point does not follow its own rotation. At ~2 rad of spin per step the solver thinks the contact is separating and under-supports the ball.

Even at lower speeds this causes serious ghost bumps.

**To reproduce:**

2D or 3D: spawn a small ball (e.g. `Collider::circle(0.008)` / `Collider::sphere(0.008)`) on a static ground plane, give it 1 m/s of rolling velocity, and watch its height relative to the ground.

Image

francisdb:
# Objective

A fast-rolling ball sinks into the ground: a 16 mm marble at 1 m/s rides 55% of its radius deep at 60 Hz. The substep separation update rotates contact anchors as material points, but a ball's contact point does not follow its own rotation. At ~2 rad of spin per step the solver thinks the contact is separating and under-supports the ball.

Even at lower speeds this causes serious ghost bumps.

## Solution

Split the ball-side anchor: only the COM-to-center part rotates, the center-to-contact part stays fixed. Stored as `anchor_pivot1/2: Option<Vector>` on `ContactManifold`/`ContactConstraint`; `None` = old behavior.

- AI disclosure: Used fore debugging my game, creating a reproducer app, help writing the fix
- Migration guide: two new public fields.

## Future work

The same error affects any rotationally symmetric shape rolling on its curved surface, e.g. a capsule or cylinder on its side. Those are only symmetric about one axis, so only the spin about that axis can be excluded, and each contact point needs its own pivot (the closest point on the inner segment) instead of the single per-manifold pivot added here. Since severity scales with spin per substep, fast-spinning balls are by far the worst case; the rest can be a follow-up if the need shows up.

## Testing

- New test `fast_rolling_ball_does_not_sink` (2D + 3D): 48% ride depth without the fix, <5% with it.
- Full suite green with the CI feature set; determinism hash unchanged.
- Ride depth, 0.008 m ball at 1 m/s:

  | tick rate | before | after |
  |-----------|--------|-------|
  | 30 Hz     | 98%    | 1.6%  |
  | 60 Hz     | 54.7%  | 0.5%  |
  | 240 Hz    | 3.5%   | 0.2%  |

---

### Before

image

### After

image

Fixes #1056

### DIFF
--- a/migration-guides/0.7-to-main.md
+++ b/migration-guides/0.7-to-main.md
@@ -94,3 +94,15 @@ consider using the `FlushContactStatusChangesQueue` command.
 Each `SolverBody` and `SolverBodyInertia` associated with awake dynamic and kinematic bodies
 is now stored in a `SolverBodies` resource instead of a component on the entity. The index
 of the solver body associated with a given entity is stored in the `SolverBodyIndex` component.
+
+## Contact Anchor Pivots
+
+`ContactManifold` and `ContactConstraint` have two new public fields,
+`anchor_pivot1` and `anchor_pivot2`. When set, only the part of the contact
+anchors from the body's center of mass to the pivot follows the body's
+rotation when tracking contact separation during substeps. This is used for
+rotationally symmetric shapes like circles and spheres to keep fast-spinning
+rolling bodies from sinking into the ground.
+
+If you construct these types manually, set the fields to `None` to keep
+the previous behavior of treating the full anchors as material points.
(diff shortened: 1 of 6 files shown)


francisdb:
@Jondolf The pending 'Test Suite' check looks like a leftover from #1016: the job is now a matrix and reports as    
  'Test Suite (ubuntu-latest)', so the old required-check name never gets a status.

francisdb:
@Jondolf any feedback would be welcome

PR #1040, opened 2026-07-31
Fix velocity_projection bench never compiling

francisdb:
# Objective

Remove invalid-feature warning on every build

## Solution

The velocity_projection bench target required a "test" feature that was never declared, so cargo skipped the bench and printed an invalid-feature warning on every build. Since the bench never compiled, it had also bit-rotted: it imported the since-removed avian3d::math::PI along with an unused Vec3.

Declare the test feature in avian2d and avian3d, gate the shared test utilities module on cfg(any(test, feature = "test")) so benches can use QuasiRandomDirection, and drop the stale bench imports. The bench now compiles and runs with: cargo bench -p avian3d --features test

I kept the `test` feature name from the bench's original `required-features`, but can rename it to something like `test-util` if you prefer, since `cfg(any(test, feature = "test"))` is confusing.

AI assisted 

## Testing

The CI clippy step now runs cargo clippy --locked --all-targets --features avian2d/test,avian3d/test, which compiles and lints the bench on every PR. Verified locally with CI's deny-warnings flags

### DIFF
--- a/.github/workflows/rust.yaml
+++ b/.github/workflows/rust.yaml
@@ -166,4 +166,4 @@ jobs:
             - name: Run cargo fmt
               run: cargo fmt --all -- --check
             - name: Run cargo clippy
-              run: cargo clippy --locked --all-targets
+              run: cargo clippy --locked --all-targets --features avian2d/test,avian3d/test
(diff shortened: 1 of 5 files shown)


francisdb:
~~The Test Suite failure is unrelated to this PR: since nightly 2026-07-30, cargo's new build-dir layout (rust-lang/cargo#17258) breaks cargo-nextest's dylib search path, so any binary needing libbevy_dylib.so fails to load; all current PRs fail the same way.~~ UPDATE: Fixed on the side of nextest

PR #1041, opened 2026-07-31
Opt out of cargo build-dir layout v2 in CI test job

francisdb:
# Objective

Since cargo nightly 2026-07-30, cargo enables its new build-dir layout by default (rust-lang/cargo#17258). cargo-nextest does not support the new layout yet: it builds the runtime library search path from the old target/debug[/deps] convention, so with bevy/dynamic_linking every test binary fails to load libbevy_dylib.so and test listing aborts. This currently breaks the Test Suite job on every PR.

## Solution

Set cargo's temporary opt-out env var `__CARGO_TEMPORARY_BUILD_DIR_NEW_LAYOUT_OPT_OUT` in the test job to keep the old layout until a nextest release supports layout v2.

AI assisted

## Testing

CI should now succeed, unlike #1040 

Reported on the nextest side as https://github.com/nextest-rs/nextest/issues/3512


francisdb:
Should no longer be needed as the issue in nextest has been fixed https://github.com/nextest-rs/nextest/releases/tag/cargo-nextest-0.9.143

PR #1044, opened 2026-08-06
Expose hit feature IDs in raycast results

jiangheng90:
## Motivation

Raycasting against triangle mesh colliders currently discards the hit
feature ID computed by parry (`RayIntersection::feature`), so there is no
way to know which triangle was hit.

One concrete use case is 3D Tiles / glTF feature metadata picking
(batch tables, EXT_mesh_features, EXT_structural_metadata): the standard
CPU picking flow used by Cesium for Unreal takes the hit **face index**
from the collision raycast, uses it to look up the `_FEATURE_ID` attribute,
and then queries the property table. I am implementing the same flow in a
[mini Cesium written in Rust with Bevy](https://community.cesium.com/t/experimenting-with-a-mini-cesium-implementation-in-rust/46753).

## Changes

All changes are gated behind a new **opt-in** cargo feature
`ray-hit-feature-id` (avian2d and avian3d). With the feature disabled,
nothing changes — no API, layout, or behavior difference.

- `RayHitData` gains a `feature: PackedFeatureId` field
  (`#[cfg(feature = "ray-hit-feature-id")]`).
- `Collider::cast_ray` is **unchanged**. Its parry call is factored into a
  private `ray_intersection` helper, reused by the new
  `Collider::cast_ray_with_feature` method that also returns the
  `PackedFeatureId` (converted from parry's `FeatureId`).
- `SpatialQuery::cast_ray_predicate` / `ray_hits_callback` (and therefore
  `cast_ray`, `ray_hits`, and the `RayCaster` component) fill in the
  feature ID when the feature is enabled.
- `PackedFeatureId` is now re-exported from the collision prelude.
- New test `raycast_returns_trimesh_face_index`: casts a ray at a
  two-triangle quad and asserts the reported face index.

## Notes

- For triangle meshes, parry reports a **back-face** hit on triangle `i`
  as `face(i + triangle_count)`; this convention is documented on
  `RayHitData::feature` and `Collider::cast_ray_with_feature`.
- Shapes that don't identify features report `PackedFeatureId::UNKNOWN`.

## Testing

- `cargo clippy -p avian3d/avian2d (--features ray-hit-feature-id) --lib -- -D warnings`
- `cargo test -p avian3d (--features ray-hit-feature-id) --lib`
  — all pass, with and without the feature.

### DIFF
--- a/crates/avian2d/Cargo.toml
+++ b/crates/avian2d/Cargo.toml
@@ -39,6 +39,10 @@ default-collider = []
 parry-f32 = ["dep:parry2d", "default-collider"]
 parry-f64 = ["f64", "dep:parry2d-f64", "default-collider"]
 
+# Adds the hit feature ID (e.g. the triangle index of trimesh colliders)
+# to `RayHitData` in raycasting.
+ray-hit-feature-id = []
+
 # Enables the XPBD constraint solver for joints.
 xpbd_joints = []
 
--- a/crates/avian3d/Cargo.toml
+++ b/crates/avian3d/Cargo.toml
@@ -40,6 +40,10 @@ default-collider = []
 parry-f32 = ["dep:parry3d", "default-collider"]
 parry-f64 = ["f64", "dep:parry3d-f64", "default-collider"]
 
+# Adds the hit feature ID (e.g. the triangle index of trimesh colliders)
+# to `RayHitData` in raycasting.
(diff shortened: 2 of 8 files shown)


jiangheng90:
I found this can be implement outside this crate so I'm no sure this pr is an good idea. I'm closing this for now.

PR #1057, opened 2026-08-28
Debug rendering for discrete spatial queries

### LINKED ISSUE #426: Add debug rendering for `shape_intersections`
As the title says, it looks like currently no debug rendering is done for the colliders in [`SpatialQuery::shape_intersections`](https://docs.rs/avian2d/latest/avian2d/spatial_query/struct.SpatialQuery.html#method.shape_intersections). From what I could tell reading the debug rendering code, it only renders `Collider`s findable through a query, and the ones used in `shape_intersections` aren't part of that. Currently I do have a workaround in our code by mirroring the signature, but it would be nice to have present natively. Thanks.

<details>
<summary> Code we are currently using </summary>

```rust
use bevy::color::palettes::css;
use bevy::prelude::*;
use avian2d::prelude::*;
use bevy_ecs::system::SystemParam;

#[derive(SystemParam)]
pub struct ShapeIntersections<'w, 's> {
    pub spatial_query: SpatialQuery<'w, 's>,
    pub gizmos: Gizmos<'w, 's>,
}

impl ShapeIntersections<'_, '_> {
    pub fn shape_intersections(
        &mut self,
        shape: &Collider,
        shape_position: avian2d::math::Vector,
        shape_rotation: avian2d::math::Scalar,
        query_filter: SpatialQueryFilter,
    ) -> Vec<Entity> {
        if let Some(cuboid) = shape.shape().downcast_ref::<avian2d::parry::shape::Cuboid>() {
            self.gizmos.primitive_2d(&Rectangle { half_size: cuboid.half_extents.into()}, shape_position, shape_rotation, css::YELLOW)
        };
        self.spatial_query.shape_intersections(shape, shape_position, shape_rotation, query_filter)
    }
}
```

</details>

justDeeevin:
# Objective

While #186 added debug rendering for continuous raycasts and shapecasts, other spatial queries are not rendered. The PR's description mentions that users could use a `PhysicsDebugRenderer` struct to accomplish this, but this struct seems to have since been removed, meaning the library now provides no direct means of visualizing discrete spatial queries (that is, methods called on the `SpatialQuery` struct).

Closes  #426.

## Solution

This PR adds debug rendering to the following `SpatialQuery` methods:
- `cast_ray`
- `ray_hits`
- `cast_shape`
- `shape_hits`
- `project_point`
- `shape_intersections`

This is accomplished by adding `Commands` and `GizmoConfigStore` to `SpatialQuery`'s fields and spawning entities with unique components that despawn after a configurable amount of time. These entities are not spawned if `PhysicsGizmos` is configured so as to not render them, and the code doesn't even run if the feature is not enabled. Raycasts and Shapecasts leverage the existing debug rendering code for `RayCaster` and `ShapeCaster` by spawning the respective entity alongside precomputed `RayHits` and `ShapeHits`. In order to prevent the caster components from clearing the hits components, some simple checks for the presence of the new `TempGizmo` component are added to their logic. Additionally, in order to prevent a feedback loop, the casters now directly call the `cast_*_predicate` functions each update.

## Testing

This PR also adds an example for debug rendering of 3d spatial queries which demonstrates the functionality of the additions.

---

## Showcase

You can run the above-mentioned example with `cargo run --example debug_render_3d`.

## Notes
An important side-effect of this change is that the aforementioned methods now take a mutable reference to self. However, all of these methods are actually convenience wrappers around lower-level `_predicate` and `_callback` methods, which I have deliberately left unchanged to allow an escape hatch for those who may need it (this is already being used by the PR itself as described above). I believe this is an acceptable compromise. One problem, though, is that the methods still take `&mut self` even when the `debug-plugin` feature is disabled, which is unnecessary. I don't know how this could be feasibly feature-gated, though it may be possible.

I also have no clue what colors should be used for `project_point` and `shape_intersections`. Right now, it's all just white as a placeholder.

### DIFF
--- a/src/debug_render/configuration.rs
+++ b/src/debug_render/configuration.rs
@@ -1,5 +1,6 @@
 use crate::prelude::*;
 use bevy::{color::palettes::css::*, prelude::*};
+use core::time::Duration;
 
 /// Gizmos used for debug rendering physics. See [`PhysicsDebugPlugin`]
 ///
@@ -92,6 +93,17 @@ pub struct PhysicsGizmos {
     pub shapecast_point_color: Option<Color>,
     /// The color used for the hit normals in [shapecasts](spatial_query#shapecasting).
     pub shapecast_normal_color: Option<Color>,
+    /// The color used for the origin in [point projections](spatial_query#point-projection).
+    pub point_projection_origin_color: Option<Color>,
+    /// The color used for the arrow in [point projections](spatial_query#point-projection).
+    pub point_projection_arrow_color: Option<Color>,
+    /// The color used for the projected point in [point projections](spatial_query#point-projection).
+    pub point_projection_point_color: Option<Color>,
+    /// The color used for the shapes in [shape intersections](SpatialQuery::shape_intersections).
+    pub shape_intersection_shape_color: Option<Color>,
+    /// The color used for the intersecting shapes in
+    /// [shape intersections](SpatialQuery::shape_intersections).
+    pub shape_intersection_hit_color: Option<Color>,
     /// The color used for the bounds of [`PhysicsIsland`](dynamics::solver::islands::PhysicsIsland)s.
     pub island_color: Option<Color>,
     /// The color used for [`ColliderTree`](crate::collider_tree) nodes.
@@ -99,6 +111,9 @@ pub struct PhysicsGizmos {
     /// Determines if the visibility of entities with [colliders](Collider) should be set to `Visibility::Hidden`,
     /// which will only show the debug renders.
     pub hide_meshes: bool,
+    /// How long discrete spatial queries (that is, methods called on [`SpatialQuery`]) should be
+    /// rendered.
+    pub temp_gizmo_lifetime: Duration,
 }
 
 impl Default for PhysicsGizmos {
@@ -122,6 +137,14 @@ impl Default for PhysicsGizmos {
             shapecast_shape_color: Some(Color::srgb(0.4, 0.6, 1.0)),
             shapecast_point_color: Some(YELLOW.into()),
             shapecast_normal_color: Some(PINK.into()),
+            // TODO: decide on these
+            point_projection_origin_color: Some(Color::WHITE),
+            point_projection_arrow_color: Some(Color::WHITE),
+            point_projection_point_color: Some(Color::WHITE),
+            shape_intersection_shape_color: Some(Color::WHITE),
+            shape_intersection_hit_color: Some(Color::WHITE),
+            temp_gizmo_lifetime: Duration::from_secs(2),
+            // ---------------------
             island_color: None,
             collider_tree_color: None,
             hide_meshes: false,
@@ -169,6 +192,14 @@ impl PhysicsGizmos {
             shapecast_shape_color: Some(Color::srgb(0.4, 0.6, 1.0)),
             shapecast_point_color: Some(YELLOW.into()),
             shapecast_normal_color: Some(PINK.into()),
+            // TODO: decide on these
+            point_projection_origin_color: Some(Color::WHITE),
+            point_projection_arrow_color: Some(Color::WHITE),
+            point_projection_point_color: Some(Color::WHITE),
+            shape_intersection_shape_color: Some(Color::WHITE),
+            shape_intersection_hit_color: Some(Color::WHITE),
+            temp_gizmo_lifetime: Duration::from_secs(2),
+            // ---------------------
             island_color: Some(RED.into()),
             collider_tree_color: Some(Color::WHITE),
             hide_meshes: true,
@@ -198,6 +229,12 @@ impl PhysicsGizmos {
             shapecast_shape_color: None,
             shapecast_point_color: None,
             shapecast_normal_color: None,
+            point_projection_origin_color: None,
+            point_projection_arrow_color: None,
+            point_projection_point_color: None,
+            shape_intersection_shape_color: None,
+            shape_intersection_hit_color: None,
+            temp_gizmo_lifetime: Duration::ZERO,
             island_color: None,
             collider_tree_color: None,
             hide_meshes: false,
--- a/src/spatial_query/system_param.rs
+++ b/src/spatial_query/system_param.rs
@@ -61,6 +61,10 @@ pub struct SpatialQuery<'w, 's> {
     colliders: Query<'w, 's, (&'static Position, &'static Rotation, &'static Collider)>,
     aabbs: Query<'w, 's, &'static ColliderAabb>,
     collider_trees: Res<'w, ColliderTrees>,
+    #[cfg(feature = "debug-plugin")]
+    commands: Commands<'w, 's>,
+    #[cfg(feature = "debug-plugin")]
+    store: Res<'w, GizmoConfigStore>,
 }
 
 impl SpatialQuery<'_, '_> {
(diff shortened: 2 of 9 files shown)


MeGaGiGaGon:
One potential footgun I see with this implementation is that there are two ways you can use a `SpatialQuery`:
- A one-off one time check
- A continuous multi-frame check

If doing a one-off check, this would be an ok implementation, but if doing a multi-frame check, you will end up with `framerate * temp gizmo lifetime` entities. This is even worse if the continuous `SpatialQuery` moves over time, since you'd end up with a two second long smear effect on the debug, which is probably not what you want.

I'd hope to see this at least mentioned somewhere in the docs, since I'm not sure there's a good way to resolve the different needs of the two use cases.

justDeeevin:
I'd argue that it'd be more idiomatic to use your own temporary `RayCaster` element for a multi-frame check like you describe. However, if the discrete methods were used multiple frames in a row, then the user could either opt out of debug rendering with the lower-level methods or decrease the duration of the temporary gizmos.

It would be great if the duration could be configured on a per-method-call basis, but I can't think of any realistic way to do that.

justDeeevin:
Lint check failing on unrelated code

src/debug_render/configuration.rs
```diff
@@ -92,13 +93,27 @@ pub struct PhysicsGizmos {
...
     /// The color used for the bounds of [`PhysicsIsland`](dynamics::solver::islands::PhysicsIsland)s.
     pub island_color: Option<Color>,
     /// The color used for [`ColliderTree`](crate::collider_tree) nodes.
     pub collider_tree_color: Option<Color>,
     /// Determines if the visibility of entities with [colliders](Collider) should be set to `Visibility::Hidden`,
     /// which will only show the debug renders.
     pub hide_meshes: bool,
+    /// How long discrete spatial queries (that is, methods called on [`SpatialQuery`]) should be
+    /// rendered.
+    pub temp_gizmo_lifetime: Duration,
```
Jondolf, line 116: I'm not sure I like having the gizmos persist, especially at a global level like this. The majority of spatial queries via `SpatialQuery` are "immediate" in nature, and often called every frame with different inputs, for example character controller sweeps and checks, raycast-based vehicles, obstacle avoidance, camera rigs, projectile trajectories, enemy visibility checks, and so on. These should all render only on the current/next frame. People who need more control over how long and in what way their queries are debug rendered (e.g. some people could even want a fade-in-out effect) can implement those gizmos themselves at the call site.

src/spatial_query/system_param.rs
```diff
@@ -109,14 +119,39 @@ impl SpatialQuery<'_, '_> {
     /// - [`SpatialQuery::ray_hits`]
     /// - [`SpatialQuery::ray_hits_callback`]
     pub fn cast_ray(
-        &self,
+        &mut self,
```
Jondolf, line 122: I would strongly prefer if we didn't require `&mut self`, as this prevents using spatial queries in parallel contexts like `par_iter` closures, and also requires people to use `mut` for the system parameter. These APIs need to support people performing potentially thousands or tens of thousands of spatial queries each frame, and allowing parallelism is important here.

The way I would approach this is adding a `SpatialQueryGizmoBuffers` resource, which stores thread-local queues as `Parallel<Vec<SpatialQueryGizmo>>`, where `SpatialQueryGizmo` is some description of the data each spatial query method needs for rendering (relevant input + hit results, this could be an enum or something). The buffer would then be drained in `PostUpdate` in some system that draws the gizmos. This avoids locks, command dispatch overhead, and lets the methods keep using `&self`.

src/spatial_query/system_param.rs
```diff
@@ -109,14 +119,39 @@ impl SpatialQuery<'_, '_> {
...
+
+            if *enabled && !temp_gizmo_lifetime.is_zero() {
+                self.commands.spawn((
+                    RayCaster::new(origin, direction)
+                        .with_max_distance(max_distance)
+                        .with_solidness(solid)
+                        .with_query_filter(filter.clone()),
+                    RayHits(out.iter().cloned().collect()),
+                    TempGizmo::new(*temp_gizmo_lifetime),
+                ));
```
Jondolf, line 150: Commands and spawning new entities have a non-trivial cost, this is not something we can do for such a perf-sensitive feature. Also see my earlier comment, the approach I described there would avoid the need for this stuff.

Here you're also computing `out` once, but then won't the `RayCaster` also perform its own ray casts for the duration it's alive? So this single ray cast call ends up actually doing multiple. Also I'm pretty sure the results of the `RayCaster` itself will end up overwriting what you computed here for the debug rendering, so you could get a visual mismatch between what you queried and what is rendered.

src/debug_render/configuration.rs
Jondolf: Side note, we could support configuring this on a per-call basis by just having some `.debug()` method on `SpatialQuery` that returns something like a `SpatialQueryDebug` wrapper that holds an additional debug rendering context that you can configure with methods like `.with_lifetime(lifetime)` before you actually call the spatial query method. But yeah, I still personally wouldn't add built-in gizmo lifetimes for spatial queries, at least in the initial PR, we could revisit it later if it has enough demand

Jondolf (commented):
Broadly speaking I think this would be a useful feature to support, but the implementation needs a different approach to make this more seamless and avoid performance regressions. Large-scale games may need potentially thousands or even tens of thousands of spatial queries per frame, so perf is quite important here.

I left a comment describing this in more detail, along with a potential approach that I would try

src/spatial_query/system_param.rs
justDeeevin: I changed it so that the system that performs raycasts for the `RayCaster` component uses a `Without<TempGizmo>` filter to avoid redundant raycasts and possibly overriding the hits from the initial function call.

src/spatial_query/system_param.rs
justDeeevin: See, this is the kind of thing I was hoping to hear about when I made my initial comment on the linked issue. I totally forgot that `&mut` gets in the way of potential parallelism.

Your thread-local idea seems promising

src/spatial_query/system_param.rs
justDeeevin: Also, the commands and spawning only happen when the plugin is enabled and the gizmo lifetime isn't zero. This means that the performance cost is only paid if necessary

Of course this still doesn't address your overall concern about the performance of the debug behavior itself...

src/debug_render/configuration.rs
justDeeevin: A major goal I had when writing this pr was to follow other debug rendering behavior by being solely enabled/disabled by the plugin and gizmo config. I think that requiring some sort of additional code to enable debug rendering for a single function call erodes the ease of use that's present with the current system. I really like that I don't have to change any code to get debug rendering.

justDeeevin:
Thank you for the feedback! I understand your concerns. I'll look into the thread-local resource idea you mentioned.

Should I make a separate v2 pr or just force-push to this one?

src/debug_render/configuration.rs
Jondolf: Yup agreed, having it just work by default is good :)

Jondolf:
I'm cool with either, whatever you think is easiest :)

Thanks for working on this btw and sorry I didn't get to this earlier 😅 This is something I've been wanting, should be a good addition as long as perf isn't meaningfully affected

justDeeevin:
I'm gonna work on a V2

PR #1059, opened 2026-09-02
Fix clippy chunks_exact_to_as_chunks lint in mesh index extraction

francisdb:
Rust 1.98 clippy denies `chunks_exact` with a constant chunk size.
Use `as_chunks::<3>()` instead.


Jondolf:
Done in #1060

PR #1060, opened 2026-09-02
Pin CI nightly toolchain to work around rustc memory regression

francisdb:
# Objective

The test suite job has been failing on every PR since late August with exit code 143 while compiling `bevy_core_pipeline`. Recent nightlies need several times more memory to compile `bevy_render` and `bevy_pbr` (rust-lang/rust#161748), which kills the 16 GB runners.

## Solution

Pin the test job to `nightly-2026-08-16`, the last nightly that passed CI. Unpin once a Bevy release includes bevyengine/bevy#25512.

Also includes the `clippy::chunks_exact_to_as_chunks` fix from #1059 so the Lints job passes here as well; #1059 can be closed if this merges first.

## Testing

Measured peak rustc memory locally with the CI command and flags: `bevy_render` goes from 2.5 GB on the pinned nightly to 16.7 GB on the current one, independent of `-Zthreads`.

### DIFF
--- a/.github/workflows/rust.yaml
+++ b/.github/workflows/rust.yaml
@@ -96,8 +96,13 @@ jobs:
         timeout-minutes: 60
         steps:
             - uses: actions/checkout@v6
-            - uses: dtolnay/rust-toolchain@nightly
+            # Pinned: nightlies since late August 2026 need several times more memory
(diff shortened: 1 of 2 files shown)


Jondolf (approved):
Thanks! I'll go ahead and merge this and close #1059

PR #1061, opened 2026-09-06
V2: Debug rendering for `SpatialQuery`

### LINKED ISSUE #426: Add debug rendering for `shape_intersections`

justDeeevin:
# Objective

While #186 added debug rendering for continuous raycasts and shapecasts, other spatial queries are not rendered. The PR's description mentions that users could use a `PhysicsDebugRenderer` struct to accomplish this, but this struct seems to have since been removed, meaning the library now provides no direct means of visualizing discrete spatial queries (that is, methods called on the `SpatialQuery` struct).

Closes #426.

## Solution

This PR adds debug rendering to the following `SpatialQuery` methods:

- `cast_ray`
- `ray_hits`
- `cast_shape`
- `shape_hits`
- `project_point`
- `shape_intersections`

This is accomplished by leveraging Bevy's `Parallel` type in a resource to create multiple thread-local queues of spatial queries to be drawn. The queues are drained together and the gizmos are drawn in a single `PostUpdate` system. Each gizmo is only drawn for a single frame—this behavior could be changed or even configurable in the future if so desired.

The thread-local approach allows the relevant methods to gain this behavior without having to take `&mut self`. This creates two major improvements upon #1057:
1. There is no API breakage
2. The methods are still able to be run in parallel

In order to prevent redundant debug rendering, the continuous `RayCaster` and `ShapeCaster` methods now use the `_predicate` and `_callback` methods on the `SpatialQuery` parameter.

## Testing

This PR also adds an example for debug rendering of 3d spatial queries which demonstrates the functionality of the additions.

---

## Showcase

You can run the above-mentioned example with cargo run --example debug_render_3d.

### DIFF
--- a/crates/avian3d/Cargo.toml
+++ b/crates/avian3d/Cargo.toml
@@ -205,6 +205,10 @@ required-features = ["3d", "default-collider", "diagnostic_ui"]
 name = "debugdump_3d"
 required-features = ["3d", "debug-plugin"]
 
+[[example]]
+name = "debug_render_3d"
+required-features = ["3d", "debug-plugin", "bevy_scene"]
+
 [[bench]]
 name = "velocity_projection"
 required-features = ["3d", "default-collider", "test"]
--- a/crates/avian3d/examples/debug_render_3d.rs
+++ b/crates/avian3d/examples/debug_render_3d.rs
@@ -1,0 +1,86 @@
+use std::time::Duration;
+
+use avian3d::{math::RVector, prelude::*};
+use bevy::prelude::*;
+
+fn main() {
+    App::default()
+        .add_plugins((
+            DefaultPlugins,
+            PhysicsPlugins::default(),
+            PhysicsDebugPlugin,
+        ))
+        .add_systems(Startup, scene.spawn())
+        .insert_resource(Cycle::new(Duration::from_millis(500)))
+        .add_systems(Update, (cast, move_wall))
+        .run();
+}
+
+fn scene() -> impl SceneList {
+    bsn_list![
+        (
+            Camera3d
+            template_value(Transform::from_xyz(10.0, 10.0, 10.0).looking_at(Vec3::ZERO, Vec3::Y))
+        ),
+        (
+            Mesh3d(asset_value(Cuboid::new(8.0, 1.0, 1.0)))
+            MeshMaterial3d::<StandardMaterial>(asset_value(Color::WHITE))
+            Collider::cuboid(8.0, 1.0, 1.0)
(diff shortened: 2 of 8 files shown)


justDeeevin:
The default values of the new configurable colors will need to be decided upon; not really my forté

justDeeevin:
again, lint check is failing on unrelated code

justDeeevin:
just fixed an oversight in how the `Parallel` resource is accessed by `SpatialQuery` that led to a crash when the debug plugin wasn't enabled

justDeeevin:
whoops that same problem also applies to `GizmoConfigStore` apparently

Jondolf (approved):
Hi! Sorry, I ended up being super busy so it took a while to get to this again 😅

I pushed a couple of commits to clean up and simplify some things and add a bit more docs. I also picked some default colors, and fixed an issue where the shape intersection gizmo was being rendered as its own shape on top of the collider's own debug rendering, which was causing some z-fighting.

This looks good to me now, thank you for your work and patience on this!

PR #1064, opened 2026-09-08
Fallback to scale_dyn for unknown custom shapes in scale_shape

0xbeefd1ed:
# Objective
To have `scale_shape` supporting custom parry shapes.

Avian's scale_shape handles its own shape types, but for a custom parry shape it returned Unsupported unconditionally, never calling the shape's own scale_dyn. 

## Solution

In the local avian fork, the custom-shape branch of scale_shape now falls back to Shape::scale_dyn and only reports unsupported when that returns None.

## Testing

Tested to support custom shapes in my custom engine.


Jondolf (approved):
Thanks!

